PR #3, opened 2017-02-21
Draft of motor and encoder API

Zabot:
I had a couple phidgets lying around that I had been using on a robot doing SLAM, and I figured I'd wrap the API for them. In theory the macros I defined for `MotorController` should generalize to any of the phidget modules. Several of the phidget boards have common elements (digital inputs, encoder inputs, etc.) that might be worth creating interfaces for to try and minimize the amount of redundant code.

I `assert`ed that none of the `libphidget` calls would fail, but it might also be worth implementing a `PhidgetException` to throw in case of library function failures.

### DIFF
--- a/phidgets_api/include/phidgets_api/encoder.h
+++ b/phidgets_api/include/phidgets_api/encoder.h
@@ -1,0 +1,59 @@
+#ifndef PHIDGETS_API_ENCODER_H
+#define PHIDGETS_API_ENCODER_H
+
+#include "phidgets_api/phidget.h"
+
+namespace phidgets {
+
+class Encoder: public Phidget
+{
+  public:
+    Encoder();
+
+    /**@brief Gets the number of digital input channels supported by this board */
+    int getInputCount();
+
+    /** @brief Reads the current state of a digital input
+     * @param index The index of the input to read */
+    bool getInputState(int index);
+
+    /**@brief Gets the number of encoder input channels supported by this board */
+    int getEncoderCount();
+
+    /** @brief Reads the current position of an encoder
+     * @param index The index of the encoder to read */
+    int getPosition(int index);
+
+    /** @brief Sets the offset of an encoder such that current position is the specified value
+     * @param index The index of the encoder to set
+     * @param position The new value that should be returned by 'getPosition(index)' at the current position of the encoder*/
+    void setPosition(int index, int position);
+
+    /** @brief Gets the position of an encoder the last time an index pulse occured. An index pulse in this context refers to an input
+     * from the encoder the pulses high once every revolution.
+     * @param index The index of the encoder to read */
+    int getIndexPosition(int index);
+
+    /** @brief Checks if an encoder is powered on and receiving events
+     * @param index The index of the encoder to check */
+    bool getEnabled(int index);
+
+    /** @brief Set the powered state of an encoder. If an encoder is not enabled, it will not be given
+     * power, and events and changes in position will not be captured.
+     * @param index The index of the encoder to change
+     * @param enabled The new powered state of the encoder*/
+    void setEnabled(int index, bool enabled);
+
+    /* TODO Add set event handlers from phidget api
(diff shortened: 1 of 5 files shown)


mintar:
Hi @Zabot ,

thanks for your pull request! Looks good to me and I'd be ready to merge this. Replacing the asserts with exceptions also sounds like a good idea. Is there a reason why you didn't implement the event handlers yet? Are you planning on providing ROS wrappers that use the new API calls?

BTW, I'm trying to get new contributions to follow the [ROS style guide](http://wiki.ros.org/CppStyleGuide). The easiest way is to let `astyle` autoformat it for you:

```bash
sudo apt-get install astyle
astyle include/phidgets_api/encoder.h include/phidgets_api/motor.h src/encoder.cpp src/motor.cpp -n --style=allman --indent=spaces=2 --pad-oper --unpad-paren --pad-header --convert-tabs
```

mintar:
> I left the event handlers out because I wanted to make sure they were consistent with the rest of the event handlers, but since there aren't any user event handers implemented yet, I didn't implement them. They could either be direct wrappers of the C libary calls, or they could do a little more in terms of allowing for multiple listeners to be bound to one event and supporting C++ lambdas.

The existing devices do use event handlers. For example, there's `CPhidgetSpatial_set_OnSpatialData_Handler` in [imu.cpp:19](https://github.com/ros-drivers/phidgets_drivers/blob/31ba59ea34bd0a31ffbe8b942016c63c424f1b51/phidgets_api/src/imu.cpp#L19). It registers a data handler that ends up in [Imu::dataHandler](https://github.com/ros-drivers/phidgets_drivers/blob/31ba59ea34bd0a31ffbe8b942016c63c424f1b51/phidgets_api/src/imu.cpp#L39-L42), which is then implemented in the [ROS wrapper](https://github.com/ros-drivers/phidgets_drivers/blob/31ba59ea34bd0a31ffbe8b942016c63c424f1b51/phidgets_imu/src/imu_ros_i.cpp#L267-L271). I would do it exactly the same way for the new devices.

> as for the ROS wrappers, that's probably out of the scope of this PR, but it shouldn't be too hard.

Yes, it's out of scope, and this PR can be merged without them. I was just wondering if you had any code already written and lined up for future PRs. If I may ask: without any code calling the files in this PR, they don't do anything; so how did you come up with this PR? (Not complaining, just curious!)

mintar:
Oh, that sounds elegant. Do you have a link? (Just for my own curiosity, it indeed doesn't belong into this repo).

mintar:
Merged via f1ca554 and a36fb7f. Many thanks!

phidgets_api/include/phidgets_api/encoder.h
```diff
@@ -0,0 +1,59 @@
...
+     * power, and events and changes in position will not be captured.
+     * @param index The index of the encoder to change
+     * @param enabled The new powered state of the encoder*/
+    void setEnabled(int index, bool enabled);
+
+    /* TODO Add set event handlers from phidget api
+     * int CPhidgetEncoder_set_OnInputChange_Handler (CPhidgetEncoderHandle phid, int(*fptr)(CPhidgetEncoderHandle phid, void *userPtr, int index, int inputState), void *userPtr)
+     * int CPhidgetEncoder_set_OnPositionChange_Handler (CPhidgetEncoderHandle phid, int(*fptr)(CPhidgetEncoderHandle phid, void *userPtr, int index, int time, int positionChange), void *userPtr)
+     * int CPhidgetEncoder_set_OnIndex_Handler (CPhidgetEncoderHandle phid, int(*fptr)(CPhidgetEncoderHandle phid, void *userPtr, int index, int indexPosition), void *userPtr)
+     */
```
jlblancoc, line 51: What's the meaning of these "TO-DO"? As far as I see, those handlers are already set in the ctor, and virtual methods provided for a derived class to override them... (??)

mintar: Yes, I've already made the changes addressing those TODOs in cbf4301.

jlblancoc: :+1: I loved to see & use all those private/protected methods. Really useful... thanks!

PR #7, opened 2017-04-26
phidgets_imu: add optional serial number parameter

jovobe:
This PR adds an optional parameter serial number to the IMU node. If the parameter is set the node will wait for the IMU with the given serial number. If no serial number is given it will use the first IMU which is attached.


mintar:
Thanks a lot, and sorry for not merging earlier! :)

PR #12, opened 2017-06-29
fixed broken url to phidgets library

unklar:
Phidgets changed the path to the library. Propably because they released Phidget22 driver.


mintar:
Thanks a lot! Merged via 6ce2916 .

PR #20, opened 2017-11-15
Phidgets ik

russelhowe:
Add support for the Phidgets InterfaceKit

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Drivers for the Phidgets devices. This Catkin metapackage includes:
 
   - `phidgets_imu`
   - `phidgets_high_speed_encoder`
+  - `phidgets_ik`
 
 Installing
 ---------------------------------------------
--- a/phidgets_api/CMakeLists.txt
+++ b/phidgets_api/CMakeLists.txt
@@ -15,6 +15,7 @@ include_directories(include ${catkin_INCLUDE_DIRS})
 
 add_library(phidgets_api src/encoder.cpp
                          src/imu.cpp
+                         src/ik.cpp
                          src/ir.cpp
                          src/motor.cpp
                          src/phidget.cpp)
--- a/phidgets_api/include/phidgets_api/ik.h
+++ b/phidgets_api/include/phidgets_api/ik.h
@@ -12,26 +12,26 @@ class IK: public Phidget
     IK();
 
   protected:
- 
+
     CPhidgetInterfaceKitHandle ik_handle_;
     virtual void sensorHandler(
-      int index, 
+      int index,
       int sensorValue);
     virtual void inputHandler(
-      int index, 
+      int index,
       int inputValue);
 
   private:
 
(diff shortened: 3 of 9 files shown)


mintar:
Looks great! I'll review this next week, can't do it before. Thanks for the PR!

jlblancoc:
👍 great, thanks!! A really useful new node.

Just a couple of short comments after a quick look around:
* Would you consider changing the names "sensor" to "ADC" or "AnalogInput" or alike? I think their meaning will be clearer like that.
* Sensor values are published as a "double" between [0,1], whereas they are Volts... why don't multiplying them by "5.0" and publish real voltage values?

dogoepp:
Thanks, it's great ! I tested it with our PhidgetInterfaceKit 0/0/4 (1014_2, control four relays).

I wonder if it would not be better suited to use services to set output states. Indeed, I'm affraid of the possible consequence of message loss. If you agree, I can take care of the modification.

mintar:
Ooops, I totally forgot about this PR and was only reminded of it by @dogoepp 's comment. I'll review as soon as possible.

russelhowe:
@dogoepp I haven't had any problems with message loss, though I understand the desire for confirmation of message receipt. I like having a topic, but it seems there is no reason you couldn't have both a topic and a service to set the output.

mintar:
I've cherry-picked most commits from this PR and merged them into kinetic and lunar. I haven't cherry-picked the following commits:

* "Silence unused argument warning." (f1f6dd11)
    
That commit indeed silences the "argument unused" warning, but makes the code harder to read, since it's not apparent why the argument is casted to (void) but then not used. I've fixed it differently in 44ecfbf0.

* "Fix compile warning about incompatible types" (e8bf7895)
* "Earlier warning fix was incomplete" (df954b23)

These commits fix the warning, but then in https://github.com/russelhowe/phidgets_drivers/blob/9c1219e34924b736e2321af7a450ba3df42e9b45/phidgets_ik/src/ik_ros_i.cpp#L70 , a value of type size_t is put into a receiver type of type int, which may not fit. I've fixed it differently in ef5b1d57.

----------------------------------------------------------------------

 I've also made a couple of other minor changes. Thanks for this nice contribution! :)

mintar:
@dogoepp : Like @russelhowe , I also haven't ever had any problems with message loss - or rather, whenever I did, it was due to an insufficient queue on the Publisher/Subscriber side, or due to publishing immediately after creating a publisher (before the subscriber had a chance to connect). In other words, it was always a bug in my code, never a message lost on the network (even though I know that ROS doesn't guarantee message delivery).

However, I agree with @russelhowe that having services in addition to topics would be a good idea in this case. So if you want, please open a new PR based on the latest kinetic branch.

dogoepp:
Thanks @mintar and @russelhowe. I did as suggested in #22. Let me know if it's fine with you.

PR #22, opened 2018-01-17
Add service to set IK output state

dogoepp:
Follows the discussion in #20.

### DIFF
--- a/phidgets_ik/src/ik_ros_i.cpp
+++ b/phidgets_ik/src/ik_ros_i.cpp
@@ -49,6 +49,7 @@ void IKRosI::initDevice()
     snprintf(topicname, sizeof(topicname), "DigitalOutput%02d", i);
     boost::shared_ptr<OutputSetter> s (new OutputSetter(ik_handle_, i));
     s->subscription = nh_.subscribe(topicname, 1, &OutputSetter::set_msg_callback, s);
+    s->service = nh_.advertiseService(topicname, &OutputSetter::set_srv_callback, s);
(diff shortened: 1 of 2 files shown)


dogoepp:
If you think it's a sensible choice, I could also make it possible for the user to choose between services or topics, for instance through a parameter.

phidgets_ik/src/ik_ros_i.cpp
```diff
@@ -49,6 +49,7 @@ void IKRosI::initDevice()
     snprintf(topicname, sizeof(topicname), "DigitalOutput%02d", i);
     boost::shared_ptr<OutputSetter> s (new OutputSetter(ik_handle_, i));
     s->subscription = nh_.subscribe(topicname, 1, &OutputSetter::set_msg_callback, s);
+    s->service = nh_.advertiseService(topicname, &OutputSetter::set_srv_callback, s);
```
mintar: The services should be called "set_digital_output00", not "DigitalOutput00". Also, (although it's a bit unrelated to this PR, but it should be changed nonetheless), the other topics should be called "digital_output00", "digital_input00", analog_input00".

mintar:
Other than the change I've suggested above, this looks good to me!

mintar:
I wouldn't add a parameter to choose between topics and services. The user can just use either and ignore the other.

dogoepp:
All right. I changed the names for the topics and services.

mintar:
Many thanks!

dogoepp:
Thank you for accepting these changes !

PR #27, opened 2018-09-18
Add parameter use_imu_time (default true)

jspricke:
Setting use_imu_time to false will disable the imu time calibration and
always use the Host time, i.e. ros::Time::now().


mintar:
Thanks!

PR #30, opened 2018-12-28
phidgets_high_speed_encoder: fix missing tick2rad values

seebq:
Was causing some strange readings from each joint, with 4x values on first port (joint0), proper readings on second (joint1), and very strange high negative numbers on 3rd and 4th.

See: https://www.phidgets.com/phorum/viewtopic.php?f=18&t=9150


jlblancoc:
Oh my! Thanks.... it should have been parentheses instead of curly braced ones. :+1:

mintar:
Thanks!

seebq:
Thanks for the quick merge!

PR #33, opened 2019-06-25
Add missing OnInputChange handler

hikinggrass:
The phidgets api is not registering the InputHandler, which needs to be registered for the inputs of the interface kit to work.


mintar:
Thank you!

PR #38, opened 2019-08-06
Also check clang style on kinetic

mintar:


mintar:
Not working because of version differences (see failed CI test), so I'm not merging this. Should be enough to check the style on melodic anyway (since even the "kinetic" branch is checked on melodic, because I couldn't find a good way to avoid this).

PR #39, opened 2019-08-06
Switch to libphidget22

### LINKED ISSUE #35: Switch to libphidget22
I'm interested in adding support for some additional Phidgets devices, namely the [VINT hub](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=643) (and some sensors behind it).  Unfortunately, the VINT hub is not supported by libphidget21, but it is supported by libphidget22.

I've done some preliminary work to see how hard it would be to switch the current code to libphidget22.  The major issue as I see it is that libphidget22 has a very different idea of how to connect to devices.  For instance, instead of having an `InterfaceKit` structure, and having the Digital Inputs and Outputs attached to that, there seems to instead be a separate `DigitalInput` and `DigitalOutput` structure.  This is more flexible, but also means that the current class structure in `phidgets_api` doesn't really match.

Assuming we want to move to libphidget22, I can see two different ways we can make this work:
1.  Keep the current class structure, "emulating" the old ways of doing things by creating multiple device handles inside the IK class, for instance.  This has the advantage of looking like the current structure, but trades off a bunch of flexibility.
1.  Change the class structure to more closely match how libphidget22 works.  This has the advantage of flexibility, but it means that all of the old APIs, nodes, launch files, etc. would change.

Ultimately I want to port this entire thing to ROS 2, so backwards compatibility isn't a huge concern to me.  Thus I'd personally choose option 2, as it gives us access to larger numbers of devices.  However, I'd like to get some input/advice before moving forward, since I'd like to submit whatever I do to this repository.  Thanks in advance for your thoughts.

clalancette:
This is a very large PR to switch to the entire phidgets_drivers to use libphidget22.  Besides being the currently supported version, this version of the library allows us to use [VINT hub](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=643) devices, which we cannot do with libphidget21.  With this PR, I was able to test out the following phidgets (some of them on a VINT hub):

1.  [Spatial](https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=1025)
1.  [Analog Input](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=1021)
1.  [Digital Input](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=1021)
1.  [Digital Output](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=1021)
1.  [DCMotor](https://www.phidgets.com/?tier=3&catid=18&pcid=15&prodid=1089)
1.  [Temperature](https://www.phidgets.com/?tier=3&catid=14&pcid=12&prodid=724)
1.  [Thermocouple](https://www.phidgets.com/?tier=3&catid=14&pcid=12&prodid=725)

There are quite a few design decisions made in here that we need to agree on, so bear with me through each of the points below.

1.  I decided that the design of the Phidget{21} base class was a bit awkward, since there is a lot of back and forth between the base class and the derived classes.  Ultimately, the amount of common code between the different phidgets is relatively small, so for the libphidget22 port I just added some helper methods instead.  This seems to have worked well, but let me know your thoughts about it.

1.  In this PR, functions like "Digital Inputs" and "Digital Outputs" are separated into their own classes and ultimately their own ROS nodes/nodelets.  This follows the design of libphidget22, and also makes things more generic.  There is a big caveat with this method, though.  If you have an [IK](https://www.phidgets.com/?tier=3&catid=2&pcid=1&prodid=1021) Phidget, for instance, it has a set of Digital Inputs, Digital Outputs, and Analog Inputs.  You *cannot* simultaneously access the device from multiple processes at once, so you cannot run one ROS node for each of the functions.  You can access the device multiple times from the same process, so it does work to run one ROS nodelet for each function.  Given this, I would suggest we actually remove the ROS nodes altogether and just have a nodelet per "function".  This will reduce support problems later.  Thoughts?

1.  I've removed the (few) launch files that were here, mostly to keep things consistent between the different nodes.  I could add some back, though this will partially depend on what we decide for the point above.

1.  I've removed the use of diagnostics from the couple of nodes that used it.  This is to maintain consistency between nodes and to make my life easier when porting to ROS 2 (which doesn't have diagnostics quite yet).  We could revisit this decision and add it more consistently everywhere.

1.  Because things are now split into functionality classes, the "IK", and "IMU" nodes make less sense.  You can build their functionality by launching groups of other nodelets.  I've left them in place for now for compatibility reasons, but I'd suggest we remove them.

1.  All of the nodes have the concept of a `publish_rate`.  If this value is > 0, then a ROS timer is started and data will be published at that rate regardless of whether it is changing or not.  If `publish_rate` is <= 0, then data will only be published when it changes (i.e. when we get a callback from libphidget22).  Separately, for all of the nodes except for "Digital Input", there is also a `data_interval_ms`, which defines how often the device acquires data and calls the callback.  I think these are both useful so I don't have a specific question here, but I thought I'd point this out.

1.  I've removed any use of the `attachHandler` and `detachHandler` that would allow you to start one of these nodes, then plug the device in.  For my application this isn't needed, and it adds some complexity.  Do you think it is something generally useful?

1.  While the package.xml files here all define the license as BSD, there is no LICENSE file in the repository, and most of the files don't have a license header.  I'd like to add both of those; how do you feel about it?

1.  Documentation - Some of the caveats above will be good for consumers of this package to know.  How would you like to do documentation?  Update the README.md at the top-level?

1.  Given all of the changes above, and despite some of my work for backwards compatibility, this will be a very different package when this is merged.  I'd suggest creating a `noetic` branch on this repository and I can then re-target this PR to that branch.

Let me know your thoughts about the above points and anything else you can think of.  Like on my last PR, I suggest reviewing this commit-by-commit, as each one ports/adds a single node.

Fixes #35


clalancette:
Closing and re-opening to retrigger travis.

clalancette:
@mintar Any thoughts here?  I don't need a full review/merge at the moment, but I'd like to get some feedback on the general direction.  If you like the direction, then I can go ahead and start doing the ROS 2 port.  If you think there are significant structural/organizational changes to be done, I'll hold off.

Thanks!

mintar:
I've finally had time to review the PR. First off: Thanks, and I like the general direction of the PR. I don't think any structural changes are necessary, but perhaps it's easier to finish iterating on this PR before doing the ROS 2 port, so you don't need to port any changes forward. It's up to you.

> 1. I decided that the design of the Phidget{21} base class was a bit awkward, since there is a lot of back and forth between the base class and the derived classes. Ultimately, the amount of common code between the different phidgets is relatively small, so for the libphidget22 port I just added some helper methods instead. This seems to have worked well, but let me know your thoughts about it.

Agreed, it looks nice.

> 2. In this PR, functions like "Digital Inputs" and "Digital Outputs" are separated into their own classes and ultimately their own ROS nodes/nodelets. This follows the design of libphidget22, and also makes things more generic. There is a big caveat with this method, though. If you have an IK Phidget, for instance, it has a set of Digital Inputs, Digital Outputs, and Analog Inputs. You cannot simultaneously access the device from multiple processes at once, so you cannot run one ROS node for each of the functions. You can access the device multiple times from the same process, so it does work to run one ROS nodelet for each function. Given this, I would suggest we actually remove the ROS nodes altogether and just have a nodelet per "function". This will reduce support problems later. Thoughts?

I agree. Debugging nodelets is a PITA, so I prefer nodes where possible, but in this case I think it's better to remove the nodes.

> 3. I've removed the (few) launch files that were here, mostly to keep things consistent between the different nodes. I could add some back, though this will partially depend on what we decide for the point above.

I'd prefer to bring the launch files back (only the ones with the nodelets). They serve as examples on how to use the code for each piece of hardware. Especially if we start requiring several different nodelets to be started simultaneously for one device, it's nice to have an example launch file.

> 4. I've removed the use of diagnostics from the couple of nodes that used it. This is to maintain consistency between nodes and to make my life easier when porting to ROS 2 (which doesn't have diagnostics quite yet). We could revisit this decision and add it more consistently everywhere.

I'd prefer having the diagnostics topics back. It doesn't have to happen within this PR though, we could first merge this PR into `noetic` and add the diagnostics later.

> 5. Because things are now split into functionality classes, the "IK", and "IMU" nodes make less sense. You can build their functionality by launching groups of other nodelets. I've left them in place for now for compatibility reasons, but I'd suggest we remove them.

If there is a launch file that composes different nodelets to provide the same functionality, I'm open to removing them.

> 6. All of the nodes have the concept of a publish_rate. If this value is > 0, then a ROS timer is started and data will be published at that rate regardless of whether it is changing or not. If publish_rate is <= 0, then data will only be published when it changes (i.e. when we get a callback from libphidget22). Separately, for all of the nodes except for "Digital Input", there is also a data_interval_ms, which defines how often the device acquires data and calls the callback. I think these are both useful so I don't have a specific question here, but I thought I'd point this out.

I've only had hands-on experience with the Phidgets Spatial. There, the messages basically change continually - no two subsequent messages will be the same. In this case, I strongly prefer publishing the data as soon as we get the callback from libphidget22 - it simply makes no sense to publish stale data repeatedly, or to drop messages. We can still keep the `publish_rate` parameter, but I would prefer to make the default `0` here.

I don't know how the other sensors behave. Assuming a digital output only triggers a callback whenever it changes from "true" to "false" or vice versa, then it would make sense to periodically republish the current state (otherwise there could be a long time between messages on that topic).

> 7. I've removed any use of the attachHandler and detachHandler that would allow you to start one of these nodes, then plug the device in. For my application this isn't needed, and it adds some complexity. Do you think it is something generally useful?

For my application neither - I don't know about other users, but I assume it's okay not to support this use case, if we improve the error messages a bit. Currently the nodelet just crashes and the nodelet manager silently swallows the exception if no device is plugged in - it would be nice if the user got an error message saying that no device (with the specified serial number, if given) was found.

> 8. While the package.xml files here all define the license as BSD, there is no LICENSE file in the repository, and most of the files don't have a license header. I'd like to add both of those; how do you feel about it?

Yes, that sounds like a great idea, it makes the licensing situation more explicit. I'm not a lawyer, but a bit of googling sounded like it's okay to link BSD code to a LGPLv3 library (libphidget22).

> 9. Documentation - Some of the caveats above will be good for consumers of this package to know. How would you like to do documentation? Update the README.md at the top-level?

Definitely as markdown within this repo. It's up to you whether you want to do this all within the top-level README, or split among the packages (or in a separate doc subfolder) with links from the top-level README. Up until now, the documentation was all on the ROS wiki, but I believe it's better to move it into this repo.

> 10. Given all of the changes above, and despite some of my work for backwards compatibility, this will be a very different package when this is merged. I'd suggest creating a noetic branch on this repository and I can then re-target this PR to that branch.

Great idea. I've just created the `noetic` branch, please re-target.

I have some more in-depth comments for the Phidgets Spatial. I'll add comments about that tomorrow.

mintar:
@jlblancoc : You have a high-speed encoder, right? If you want to test your device with the new code, or if you have any comments, now is the time. :)

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+                  std::placeholders::_1, std::placeholders::_2,
+                  std::placeholders::_3, std::placeholders::_4));
+
+    ROS_INFO("Connected");
+
+    spatial_->setDataInterval(data_interval_ms);
+
+    imu_pub_ = nh_.advertise<sensor_msgs::Imu>("imu/data_raw", 1);
+
+    cal_publisher_ = nh_.advertise<std_msgs::Bool>("spatial/is_calibrated", 5);
```
mintar: The topic was called `imu/is_calibrated` before this PR, and the other topics are also under the `imu/` namespace, so I think we should keep it that way.

```suggestion
    cal_publisher_ = nh_.advertise<std_msgs::Bool>("imu/is_calibrated", 5);
```

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    calibrate();
+
+    cal_srv_ = nh_.advertiseService("spatial/calibrate",
```
mintar: See above.

```suggestion
    cal_srv_ = nh_.advertiseService("imu/calibrate",
```

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+    if (!nh_private_.getParam("magnetic_field_stdev", magnetic_field_stdev_))
+    {
+        magnetic_field_stdev_ =
+            0.095 * (M_PI / 180.0);  // 0.095°/s as per manual
+    }
+    int data_interval_ms;
+    if (!nh_private.getParam("data_interval_ms", data_interval_ms))
+    {
+        data_interval_ms = 250;
```
mintar: Let's keep the old default of 8 ms. A IMU publishing at 4 Hz is not useful in most use cases.

```suggestion
        data_interval_ms = 8;
```

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+    if (!nh_private.getParam("publish_rate", publish_rate_))
+    {
+        publish_rate_ = 5;
```
mintar: The IMU data should be published exactly once, and always immediately when it is received.

```suggestion
        publish_rate_ = 0;
```

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+        "Waiting for Phidgets Spatial serial %d, hub port %d to be "
+        "attached...",
+        serial_num, hub_port);
+
+    // We take the mutex here and don't unlock until the end of the constructor
+    // to prevent a callback from trying to use the publisher before we are
+    // finished setting up.
+    std::lock_guard<std::mutex> lock(spatial_mutex_);
+
+    spatial_ = std::make_unique<Spatial>(
```
mintar: If "serial" is set to something invalid (e.g., "123456"), or if the device is not plugged in, we don't get very useful logging output. It's worst for the nodelet version, because the nodelet manager even silently swallows the "Failed to open device: Timed Out" exception. It would be nice if we could tell the user what is going wrong.

**nodelet:**
```
[ INFO] [1565703216.451936437]: Initializing Phidgets SPATIAL Nodelet (/imu_manager)
[ INFO] [1565703216.452050889]: Starting Phidgets SPATIAL (/imu_manager)
[ INFO] [1565703216.452072684]: Opening spatial (/imu_manager)
[ INFO] [1565703216.469385824]: Waiting for Phidgets Spatial serial 123456, hub port 0 to be attached... (/imu_manager)
[FATAL] [1565703217.471012808]: Failed to load nodelet '/PhidgetsSpatialNodelet` of type `phidgets_spatial/PhidgetsSpatialNodelet` to manager `imu_manager' (/PhidgetsSpatialNodelet)
[PhidgetsSpatialNodelet-2] process has died [pid 7133, exit code 255, cmd /opt/ros/kinetic/lib/nodelet/nodelet load phidgets_spatial/PhidgetsSpatialNodelet imu_manager __name:=PhidgetsSpatialNodelet __log:=/home/martin/.ros/log/2d7e8006-bdca-11e9-91d3-5cc5d4df8e50/PhidgetsSpatialNodelet-2.log].
log file: /home/martin/.ros/log/2d7e8006-bdca-11e9-91d3-5cc5d4df8e50/PhidgetsSpatialNodelet-2*.log
```

**node:**
```
[ INFO] [1565703303.125249565]: Starting Phidgets SPATIAL (/PhidgetsSpatialNodelet)
[ INFO] [1565703303.125295119]: Opening spatial (/PhidgetsSpatialNodelet)
[ INFO] [1565703303.144419346]: Waiting for Phidgets Spatial serial 123456, hub port 0 to be attached... (/PhidgetsSpatialNodelet)
terminate called after throwing an instance of 'phidgets::Phidget22Error'
  what():  Failed to open device: Timed Out
[PhidgetsSpatialNodelet-1] process has died [pid 7408, exit code -6, cmd /home/martin/ros-kinetic/devel/lib/phidgets_spatial/phidgets_spatial_node __name:=PhidgetsSpatialNodelet __log:=/home/martin/.ros/log/2d7e8006-bdca-11e9-91d3-5cc5d4df8e50/PhidgetsSpatialNodelet-1.log].
log file: /home/martin/.ros/log/2d7e8006-bdca-11e9-91d3-5cc5d4df8e50/PhidgetsSpatialNodelet-1*.log
```

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
...
+                msg->linear_acceleration_covariance[idx] = 0.0;
+                msg->angular_velocity_covariance[idx] = 0.0;
+            }
+        }
+    }
+
+    double imu_diff_in_secs = (last_data_timestamp_ - data_time_zero_) / 1000.0;
+    double time_in_secs = ros_time_zero_.toSec() + imu_diff_in_secs;
+
+    msg->header.stamp = ros::Time(time_in_secs);
```
mintar: The parameter `use_imu_time` that was introduced in #27 is being removed in this PR. The reason why it was introduced is that the IMU's clock is notoriously inaccurate (at least in the old Phidget Spatial 1044_0 that we have). The behavior when `use_imu_time == true`, which is also implemented here, is to believe the IMU's clock. This has the advantage of avoiding USB jitter in the timestamps (so they come at more evenly spaced intervals and don't "bunch up"), but over time, the time lag between IMU and host PC builds up. This is why `use_imu_time == true` simply uses the host PC's time.

In my opinion, one of the following things should be done:

1. Bring back `use_imu_time`, or:
2. Throw away the whole complicated time handling and simply implement the behavior of `use_imu_time == false`, that is, simply use ros::Time::now() (at the time when the message was received).

I'd probably go with (2).

phidgets_spatial/src/spatial_ros_i.cpp
```diff
@@ -0,0 +1,295 @@
+#include <functional>
```
mintar, line 30: We should probably add a "migration guide" for people upgrading to Noetic. Here are the entries for the Phidgets Spatial:

- parameters renamed:

    * `period -> data_interval_ms`
    * `serial_number -> serial`

- frame_id: default changed "imu" -> "imu_link" as specified in http://www.ros.org/reps/rep-0145.html
- diagnostics are gone (but will hopefully be added back later)

mintar (changes requested):
I've reviewed the Phidgets Spatial in depth. Everything looks good, except the proposed changes in the comments below.

I've also reviewed all other commits in this PR and they look ok, but I can't guarantee that there are no small details that would be good to change (as for the Spatial), since I don't have any other devices for testing. Since this PR is going to get merged into Noetic, and we won't break any existing installations, I'll still merge it once the comments are addressed.

phidgets_spatial/src/spatial_ros_i.cpp
mintar: Oh, and the "Waiting to be attached" is a bit misleading and should be rephrased, since we don't really wait any more (only 1 second, instead of indefinitely).

clalancette:
@mintar Fantastic, thanks for the general and detailed review!  I will start addressing your comments soon.

clalancette:
This is a reply to the general points.  I'll just remove bits that we agree on to cut down on the length.

> > I would suggest we actually remove the ROS nodes altogether and just have a nodelet per "function". This will reduce support problems later. Thoughts?
> 
> I agree. Debugging nodelets is a PITA, so I prefer nodes where possible, but in this case I think it's better to remove the nodes.

I'll go ahead and remove the nodes.

> > 1. I've removed the (few) launch files that were here, mostly to keep things consistent between the different nodes. I could add some back, though this will partially depend on what we decide for the point above.
> 
> I'd prefer to bring the launch files back (only the ones with the nodelets). They serve as examples on how to use the code for each piece of hardware. Especially if we start requiring several different nodelets to be started simultaneously for one device, it's nice to have an example launch file.

OK, I'll bring back launch files for the individual nodes (at least).  I'll also make a launch file for the IK nodelet as well, instead of having a duplicate implementation.

> > 1. I've removed the use of diagnostics from the couple of nodes that used it. This is to maintain consistency between nodes and to make my life easier when porting to ROS 2 (which doesn't have diagnostics quite yet). We could revisit this decision and add it more consistently everywhere.
> 
> I'd prefer having the diagnostics topics back. It doesn't have to happen within this PR though, we could first merge this PR into `noetic` and add the diagnostics later.

OK, I'll open an issue to track that work for the future.

> > 1. Because things are now split into functionality classes, the "IK", and "IMU" nodes make less sense. You can build their functionality by launching groups of other nodelets. I've left them in place for now for compatibility reasons, but I'd suggest we remove them.
> 
> If there is a launch file that composes different nodelets to provide the same functionality, I'm open to removing them.

Yeah, as I mentioned above, I'll probably remove the IK code and just make a launch file for it instead.  Thinking about it again, I'm somewhat tempted to leave the Spatial node in place, since it *does* use different (if mostly equivalent) code.

> > 1. All of the nodes have the concept of a publish_rate. If this value is > 0, then a ROS timer is started and data will be published at that rate regardless of whether it is changing or not. If publish_rate is <= 0, then data will only be published when it changes (i.e. when we get a callback from libphidget22). Separately, for all of the nodes except for "Digital Input", there is also a data_interval_ms, which defines how often the device acquires data and calls the callback. I think these are both useful so I don't have a specific question here, but I thought I'd point this out.
> 
> I've only had hands-on experience with the Phidgets Spatial. There, the messages basically change continually - no two subsequent messages will be the same. In this case, I strongly prefer publishing the data as soon as we get the callback from libphidget22 - it simply makes no sense to publish stale data repeatedly, or to drop messages. We can still keep the `publish_rate` parameter, but I would prefer to make the default `0` here.
> 
> I don't know how the other sensors behave. Assuming a digital output only triggers a callback whenever it changes from "true" to "false" or vice versa, then it would make sense to periodically republish the current state (otherwise there could be a long time between messages on that topic).

You hit it exactly right.  The libphidget22 callback for most of these devices fire when there is a change.  For most devices, this happens at a fixed rate, but for the digital input it only happens when they change.  That's why I added the ability to publish on a timer.  But I generally agree that we should only publish on change by default, so I'll make the default 0 for `publish_rate`.

> 
> > 1. I've removed any use of the attachHandler and detachHandler that would allow you to start one of these nodes, then plug the device in. For my application this isn't needed, and it adds some complexity. Do you think it is something generally useful?
> 
> For my application neither - I don't know about other users, but I assume it's okay not to support this use case, if we improve the error messages a bit. Currently the nodelet just crashes and the nodelet manager silently swallows the exception if no device is plugged in - it would be nice if the user got an error message saying that no device (with the specified serial number, if given) was found.

OK, I'll poke at this a bit and see if I can't get better error messages.

> 
> > 1. While the package.xml files here all define the license as BSD, there is no LICENSE file in the repository, and most of the files don't have a license header. I'd like to add both of those; how do you feel about it?
> 
> Yes, that sounds like a great idea, it makes the licensing situation more explicit. I'm not a lawyer, but a bit of googling sounded like it's okay to link BSD code to a LGPLv3 library (libphidget22).

Oh, right, I totally forgot about that bit.  My understanding is that it is fine to link BSD into LGPLv3, but the result is equivalent to LGLPv3.  I'll make sure to add some LICENSING terms that cover both of these parts.

> 
> > 1. Documentation - Some of the caveats above will be good for consumers of this package to know. How would you like to do documentation? Update the README.md at the top-level?
> 
> Definitely as markdown within this repo. It's up to you whether you want to do this all within the top-level README, or split among the packages (or in a separate doc subfolder) with links from the top-level README. Up until now, the documentation was all on the ROS wiki, but I believe it's better to move it into this repo.
> 
> > 1. Given all of the changes above, and despite some of my work for backwards compatibility, this will be a very different package when this is merged. I'd suggest creating a noetic branch on this repository and I can then re-target this PR to that branch.
> 
> Great idea. I've just created the `noetic` branch, please re-target.

I've retargetted the PR.

mintar:
I agree with all of your comments. Just two remarks:

> I'll go ahead and remove the nodes.

Perhaps it's worthwhile to only remove those nodes that would cause the support problems that we talked about earlier. For example, if somebody is using an InterfaceKit and then starts the `phidgets_digital_inputs_node` and the `phidgets_digital_outputs_node`, it won't work - they need to use the nodelets instead. On the other hand, starting the `phidgets_spatial_node` would be fine. What do you think?

> Oh, right, I totally forgot about that bit. My understanding is that it is fine to link BSD into LGPLv3, but the result is equivalent to LGLPv3. I'll make sure to add some LICENSING terms that cover both of these parts.

That would be true of GPL, but after reading the [wikipedia article](https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License) and the [license text](https://www.gnu.org/licenses/lgpl-3.0.de.html) I believe LGPL allows exactly this use case. Since we dynamically link to the `libphidget22` library, and make it possible for the user to modify the LGPL-licensed library and compile it with our code, we can put our code under any license we want, including BSD.

Therefore I believe the `libphidget22` package is LGPL and the others are BSD (exactly as the package.xml states), and it would be totally fine to add BSD license headers to all of our source files. According to [this answer](https://softwareengineering.stackexchange.com/a/371526/242820) it would even be okay to add a LICENSE file with only the BSD text in the root folder, and then maybe a separate LICENSE file with the LGPLv3 text in the `libphidget22` subfolder or something like that.

clalancette:
> Perhaps it's worthwhile to only remove those nodes that would cause the support problems that we talked about earlier. For example, if somebody is using an InterfaceKit and then starts the `phidgets_digital_inputs_node` and the `phidgets_digital_outputs_node`, it won't work - they need to use the nodelets instead. On the other hand, starting the `phidgets_spatial_node` would be fine. What do you think?

The issue is that there are Spatial phidgets that connect to a VINT hub: https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=975 .  If you have a temperature sensor also connected to the VINT hub, then you need to use the nodelet version of both to make it work.  This goes for pretty much all of the phidgets; it looks like they are moving to a world where everything is hooked to a VINT hub.

I see that there is value in supporting both nodes and nodelets; as you mentioned, the former are easier to debug.  And I honestly don't care which way we go here, I'm just trying to reduce support burden.  For now, I'll push a commit that removes all of them, but just let me know what you want to do and I'm happy to revert or modify that.

> Therefore I believe the `libphidget22` package is LGPL and the others are BSD (exactly as the package.xml states), and it would be totally fine to add BSD license headers to all of our source files. According to [this answer](https://softwareengineering.stackexchange.com/a/371526/242820) it would even be okay to add a LICENSE file with only the BSD text in the root folder, and then maybe a separate LICENSE file with the LGPLv3 text in the `libphidget22` subfolder or something like that.

Ah, good point.  I've added the BSD license headers everywhere and I've added both a LICENSE.lgplv3 and LICENSE.bsd at the top level.  I'll push that commit shortly as well.

Just a note that I'm on vacation next week (Aug 19 - 23), so for anything I don't get to I'll pick up the following week.

phidgets_spatial/src/spatial_ros_i.cpp
clalancette: Ha, yeah.  That was a `sed` error, will fix.

phidgets_spatial/src/spatial_ros_i.cpp
clalancette: Yeah, agreed it is not that useful.  I was going what was listed as the default from the API: https://www.phidgets.com/?tier=3&catid=10&pcid=8&prodid=975#product_tab_5 .  But I can switch it back too.  I'll do the same for the individual accelerometer, magnetometer, and gyroscope as well.

phidgets_spatial/src/spatial_ros_i.cpp
clalancette: Commit fdfdefb  has set the `publish_rate_` to a default of 0 for all drivers.

phidgets_spatial/src/spatial_ros_i.cpp
clalancette: Commit dfdc7e637023e4886235c9f91de66a8c8d6dcadd updates the wording to be `Connecting to ...`, which I think is a little clearer.

The only thing I can think to do about the failure case for nodelets is to put a `try/catch` block around the connection, and print out the error when it fails, and re-raise.  I'll implement that for all of the nodes.

phidgets_spatial/src/spatial_ros_i.cpp
clalancette: > The behavior when `use_imu_time == true`, which is also implemented here, is to believe the IMU's clock. This has the advantage of avoiding USB jitter in the timestamps (so they come at more evenly spaced intervals and don't "bunch up"), but over time, the time lag between IMU and host PC builds up. This is why `use_imu_time == true` simply uses the host PC's time.

Hm, right.  I've noticed the USB bunching pretty severely in other IMU devices I've worked with, which is why I think `use_imu_time==true` is better in general.

What if we periodically re-synced the clocks?  Let's say every 5 seconds, we reset `ros_time_zero` and `data_time_zero` to `ros::Time::now()` and the data timestamp, respectively, but in the in-between measurements we use the delta as calculated from the device?  That has some potential to jump backwards, but I think we can ensure we never do that.  This has some of the properties of both of the proposals.  Thoughts?

clalancette:
All right, I've done an initial pass over your first round of comments.  What I haven't accomplished yet:

1.  Reinstating the launch files.
1.  Rewriting the IK node as a launch file with nodelets.
1.  Writing the "migration guide"
1.  Figuring out the `use_imu_time` (I'll need input from you there)

I'll be working on those remaining items, but it may have to wait until I get back.  Thanks again for all of the reviews.

phidgets_spatial/src/spatial_ros_i.cpp
mintar: Sure, sounds like a good idea! (I mean periodically re-syncing the clocks).

LICENSE
```diff
@@ -0,0 +1,5 @@
+The libphidgets22 library is licensed under the GNU Lesser General Public Licence v3 or later.
+See LICENSE.gplv3 and LICENSE.lgplv3 for the text of the license.
+
+The rest of the code in the library is licensed under the 3-clause BSD license.
```
mintar:
```suggestion
The rest of the code in this repository is licensed under the 3-clause BSD license.
```

phidgets_magnetometer/README.md
```diff
@@ -0,0 +1,30 @@
...
+ * `cc_offset2` (double) - Calibration offset value 2; see device's user guide for information on how to calibrate.
+ * `cc_gain0` (double) - Gain offset value 0; see device's user guide for information on how to calibrate.
+ * `cc_gain1` (double) - Gain offset value 1; see device's user guide for information on how to calibrate.
+ * `cc_gain2` (double) - Gain offset value 2; see device's user guide for information on how to calibrate.
+ * `cc_t0` (double) - T offset value 0; see device's user guide for information on how to calibrate.
+ * `cc_t1` (double) - T offset value 1; see device's user guide for information on how to calibrate.
+ * `cc_t2` (double) - T offset value 2; see device's user guide for information on how to calibrate.
+ * `cc_t3` (double) - T offset value 3; see device's user guide for information on how to calibrate.
+ * `cc_t4` (double) - T offset value 4; see device's user guide for information on how to calibrate.
+ * `cc_t5` (double) - T offset value 5; see device's user guide for information on how to calibrate
```
mintar: Should be "Gain" and "T", not "Gain offset" and "T offset". Same for spatial.

mintar:
Thanks again for yet another round of commits! I've added two more (typo) comments. Apart from those, only the 4 points you've mentioned remain from my point. I'll also be on vacation next week. You're not by any chance visiting the chaos communication camp, are you? :)

Regarding the ROS2 port: I think it would make most sense to develop it in a separate branch in this repository. I haven't had any time to work with ROS2 yet, so it would probably make most sense if I gave you write access to this repo for working on ROS2 (I couldn't provide meaningful reviews anyway). What do you think?

clalancette:
All right, I've addressed 3 of the 4 points.  I don't have my phidgets spatial with me, so I can't test out my ideas for `use_imu_time`; I'll do that when I get back from vacation.  Other than that, everything else should be done, barring more comments.

> Thanks again for yet another round of commits! I've added two more (typo) comments. Apart from those, only the 4 points you've mentioned remain from my point. I'll also be on vacation next week. You're not by any chance visiting the chaos communication camp, are you? :)

No, family vacation.  But it sounds like fun.  Will you be at ROSCon?

> 
> Regarding the ROS2 port: I think it would make most sense to develop it in a separate branch in this repository. I haven't had any time to work with ROS2 yet, so it would probably make most sense if I gave you write access to this repo for working on ROS2 (I couldn't provide meaningful reviews anyway). What do you think?

Yeah, that makes perfect sense to me.  What I'll probably do is create a `dashing` branch, then open PRs against the `dashing` branch as I work.  That gives you visibility into what I'm doing, and gives you a chance to comment if you'd like.

mintar:
> All right, I've addressed 3 of the 4 points. I don't have my phidgets spatial with me, so I can't test out my ideas for use_imu_time; I'll do that when I get back from vacation. Other than that, everything else should be done, barring more comments.

Perfect, just comment after your vacation when it's done. I have no other comments, so we can merge then.

> No, family vacation. But it sounds like fun. Will you be at ROSCon?

Unfortunately my IROS paper was rejected, so I also won't be able to attend ROSCon. It's a shame, I was really looking forward to it.

> Yeah, that makes perfect sense to me. What I'll probably do is create a dashing branch, then open PRs against the dashing branch as I work. That gives you visibility into what I'm doing, and gives you a chance to comment if you'd like.

Sounds good! Ticketed as #45 .

clalancette:
I am back from vacation, but I've been quiet.  Sorry about that.

I have been working on the time resynchronization problem.  I ran into some snags, but I think I finally have it mostly figured out.  There is still some cleanup to do on it before I push it here, but unfortunately that will have to wait until next week.  I'll ping again when I've finished.

mintar:
OK, no worries!

clalancette:
> OK, no worries!

All right, I've finally gotten the resynchronization code done, commented, tested, and documented.  It's a rather large commit, but it is probably best to just look at the changes to `spatial_ros_i.cpp` and in particular the comment about why it looks like it does.  Let me know what you think!

mintar (approved):

mintar:
Many thanks! It's been a real pleasure working with you. Thanks for going the extra mile and implementing a proper time synchronization logic! I'm going to forward-port #41 and #46 to noetic within the next few days. Do you mind if I add you to the maintainers tag in the package.xmls?

clalancette:
> Many thanks! It's been a real pleasure working with you.

Likewise, thanks for being a helpful and responsive maintainer.

> I'm going to forward-port #41 and #46 to noetic within the next few days. 

Perfect, thanks.

> Do you mind if I add you to the maintainers tag in the package.xmls?

That is totally fine with me.

arunavanag591:
Will there be a documentation for this branch, which is currently just the kinetic documentation. Given there has been lot more support added to the noetic branch.

arunavanag591:
Will there be an update to this [wiki](https://wiki.ros.org/phidgets_api) as the noetic branch is using libphidget22?

clalancette:
> Will there be an update to this [wiki](https://wiki.ros.org/phidgets_api) as the noetic branch is using libphidget22?

Yep, as soon as we release this package for noetic, the buildfarm will pick up the README.md in here and generate the documentation for it.

arunavanag591:
Lastly, I see the motor is presently for DC motor, any future plans to include [stepper motors](https://www.phidgets.com/?tier=3&catid=23&pcid=20&prodid=1029)

clalancette:
> Lastly, I see the motor is presently for DC motor, any future plans to include [stepper motors](https://www.phidgets.com/?tier=3&catid=23&pcid=20&prodid=1029)

The open PR https://github.com/ros-drivers/phidgets_drivers/pull/56 implements this, but it needs to be cleaned up and I don't have time to work on it.  If you'd like to clean up that PR and submit a new one getting the support in, I'd be happy to review it.

Issue #42, opened 2019-08-15
Port #41 forward to noetic


mintar:
Just a small reminder to self to cherry-pick 1f6615d onto noetic once #39 is merged.

PR #47, opened 2019-09-05
Improve the IMU calibration service

mintar:
* Change misleading IMU calibration log message.
* Block 2 seconds in IMU calibration handler.
* Make is_calibrated topic latched


mintar:
This is a forward-port of #41. @clalancette , what do you think?

Fixes #42

clalancette (approved):

PR #52, opened 2019-09-24
Dashing: Various fixes after using the driver for real

clalancette:
Here's another set of fixes that I came up with as we've started to use and integrate these drivers for real:

1.  Don't fetch the analog sensor value at startup.  That's because some VoltageInput devices (like https://www.phidgets.com/?tier=3&catid=16&pcid=14&prodid=954) don't support fetching the data like that.
1.  Fix up the math for returning the read voltage data.  In the Melodic version of the code, we were getting "sensor" data, which more-or-less equates to ADC ticks.  In the new libphidget22 code, however, we are getting a direct voltage, where the underlying libphidget22 library takes care of converting the ticks to a real voltage.  Thus, we don't need to do it again.  Remove that extra code.
1.  Make sure to reset the VOLTAGE_RANGE to AUTO when connecting to the driver.  This is a belt-n-suspenders fix just intended to make sure it is in the configuration we expect.
1.  Add in a `gain` and `offset` per voltage device.  When you are using hub_port devices (i.e. non-smart devices), you may want to apply a gain and an offset to the voltage you read.  This is especially true with something like the [current sensors](https://www.phidgets.com/?tier=3&catid=16&pcid=14&prodid=438), which have a very definite formula of what the gain and offset should be to get reasonable values.
1.  Change `publish_rate` to a double.  This is so we can support fractional times like 0.5 (publish once every 2 seconds).

As usual, any comments or thoughts welcome.  @IanTheEngineer, FYI.


clalancette:
I'm going to merge this up to keep moving forward, but I'm happy to address any comments that come up from this.

PR #53, opened 2019-09-24
Only publish motor_back_emf if supported by DC Motor device

peterpolidoro:
Only publish motor_back_emf if supported by DC Motor device

### DIFF
--- a/phidgets_api/src/motor.cpp
+++ b/phidgets_api/src/motor.cpp
@@ -68,6 +68,22 @@ Motor::Motor(int32_t serial_number, int hub_port, bool is_hub_port_device,
             ret);
     }
 
+    back_emf_sensing_supported_ = true;
+    ret = PhidgetDCMotor_setBackEMFSensingState(motor_handle_, 1);
+    if (ret != EPHIDGET_OK)
+    {
+        if (ret == EPHIDGET_UNSUPPORTED)
+        {
+            back_emf_sensing_supported_ = false;
+        } else
+        {
+            throw Phidget22Error(
+                "Failed to set back EMF sensing state Motor channel " +
+                    std::to_string(channel),
+                ret);
+        }
+    }
+
     ret = PhidgetDCMotor_setOnBackEMFChangeHandler(motor_handle_,
                                                    BackEMFChangeHandler, this);
     if (ret != EPHIDGET_OK)
@@ -155,15 +171,24 @@ void Motor::setAcceleration(double acceleration) const
     }
 }
 
+bool Motor::backEMFSensingSupported() const
+{
+    return back_emf_sensing_supported_;
+}
+
 double Motor::getBackEMF() const
 {
-    double backemf;
-    PhidgetReturnCode ret = PhidgetDCMotor_getBackEMF(motor_handle_, &backemf);
-    if (ret != EPHIDGET_OK)
+    double backemf = 0.0;
+    if (back_emf_sensing_supported_)
     {
-        throw Phidget22Error("Failed to get back EMF for Motor channel " +
-                                 std::to_string(channel_),
-                             ret);
+        PhidgetReturnCode ret =
+            PhidgetDCMotor_getBackEMF(motor_handle_, &backemf);
+        if (ret != EPHIDGET_OK)
+        {
+            throw Phidget22Error("Failed to get back EMF for Motor channel " +
+                                     std::to_string(channel_),
+                                 ret);
+        }
     }
     return backemf;
 }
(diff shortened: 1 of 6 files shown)


phidgets_motors/src/motors_ros_i.cpp
```diff
@@ -101,15 +101,18 @@ MotorsRosI::MotorsRosI(const rclcpp::NodeOptions& options)
...
-            char backemftopic[] = "motor_back_emf00";
-            snprintf(backemftopic, sizeof(backemftopic), "motor_back_emf%02d",
-                     i);
-            motor_vals_[i].back_emf_pub =
-                this->create_publisher<std_msgs::msg::Float64>(backemftopic, 1);
-
             motor_vals_[i].last_duty_cycle_val = motors_->getDutyCycle(i);
-            motor_vals_[i].last_back_emf_val = motors_->getBackEMF(i);
+
+            if (motors_->backEMFSensingSupported(i))
```
clalancette, line 111: Honestly, I wouldn't bother with this.  It's somewhat subjective, but I'd rather just unconditionally create the topic.  If the underlying device doesn't support it, then we will never get a callback and never publish anything on it, which is fine.

I think you'll still want to keep the `backEMFSensingSupported` method so you can print an `RCLCPP_INFO` about whether it is supported or not.

phidgets_api/src/motor.cpp
```diff
@@ -155,15 +171,24 @@ void Motor::setAcceleration(double acceleration) const
...
+    return back_emf_sensing_supported_;
+}
+
 double Motor::getBackEMF() const
 {
-    double backemf;
-    PhidgetReturnCode ret = PhidgetDCMotor_getBackEMF(motor_handle_, &backemf);
-    if (ret != EPHIDGET_OK)
+    double backemf = 0.0;
+    if (back_emf_sensing_supported_)
```
clalancette: I think we should put this code back to how it was, and instead just add at the top:

```
if (!back_emf_sensing_supported_)
{
    throw Phidget22Error("Back EMF not supported", EPHIDGET_UNSUPPORTED);
}
```

phidgets_api/src/motor.cpp
```diff
@@ -215,7 +240,10 @@ void Motor::dutyCycleChangeHandler(double duty_cycle) const
 
 void Motor::backEMFChangeHandler(double back_emf) const
 {
-    back_emf_change_handler_(channel_, back_emf);
+    if (back_emf_sensing_supported_)
```
clalancette: This shouldn't be necessary; if back EMF is not supported, then we would have failed to install the change handler during construction.  So I think you can just remove this.

phidgets_motors/src/motors_ros_i.cpp
```diff
@@ -135,7 +138,10 @@ MotorsRosI::MotorsRosI(const rclcpp::NodeOptions& options)
         for (int i = 0; i < n_motors; ++i)
         {
             publishLatestDutyCycle(i);
-            publishLatestBackEMF(i);
+            if (motors_->backEMFSensingSupported(i))
```
clalancette: I think it is better to move this check into `publishLatestBackEMF`; that way, we avoid duplicating the logic.

clalancette (changes requested):
Thanks for the contribution!  I have a few things that I think should be changed before this gets in, but they should be relatively easy to fix.

clalancette:
I've made a few more changes and made a pull request to this pull request: https://github.com/janelia-ros/phidgets_drivers/pull/1 .  If you agree with those changes, please "Rebase and merge" that PR into this one, and then I'll merge the whole lot into the "dashing" branch.  Thanks!

peterpolidoro:
Thanks for patiently looking over these changes! Hopefully it will be easier for you to review my future changes as I learn more of your preferred coding styles. I merged, but did not rebase. Is it typical to collapse all pull request changes into a single commit? Would you like me to do that on my end or can you do that as you merge it back into the dashing branch?

clalancette:
> Thanks for patiently looking over these changes! Hopefully it will be easier for you to review my future changes as I learn more of your preferred coding styles. 

No problem at all.

> I merged, but did not rebase. Is it typical to collapse all pull request changes into a single commit?

It depends on the maintainer and the specific pull request.  In this case, I think we'll want to "Squash and rebase", since it is all one logical change.

> Would you like me to do that on my end or can you do that as you merge it back into the dashing branch?

I'll do it when I merge.  Thanks again for the contribution.

clalancette (approved):

PR #54, opened 2019-09-26
More small updates

clalancette:
This PR does a few more small updates to Phidgets:

1.  It changes the launch output to "both" so we get output to the screen and to a log.
1.  It adds in parameters to the IK launch file to serve as a useful example of how to do it.
1.  It updates the top-level README to add some more information I've learned while working with the phidgets.


clalancette:
I'm going to merge this up as well.  As before, I'm happy to respond to any comments post-merge.

Issue #55, opened 2019-09-26
Enhancement: Questions about adding support for Stepper Phidgets


peterpolidoro:
@clalancette I am working on adding the interface to stepper phidgets. I have a few questions about the implementation choices you prefer.

I created a Stepper class in the phidgets_api and I am modeling it after your Motor class. You added a Motors class since there are some phidgets that control multiple dc motors from one VINT port. There do not seem to be any stepper phidgets that control multiple motors from one port. Should I go ahead and create a Steppers class as well even though it is not really necessary just so it better matches the way dc motors are handled? Is the idea with these drivers that there will be one node per VINT port? An alternative might be to use a vector of serial numbers and hub ports when creating the Steppers class so that multiple steppers are controlled by one node, but perhaps just creating separate nodes, one for each stepper, is cleaner.

I want to expose almost all of the stepper phidget functions so I have them available for my future projects. The stepper phidget API has like 30-40 functions. Should I create one topic per function or does that make a mess of the topic namespace? I could group some together so we can set things like acceleration, control mode, and current limits in one setup topic and leave setting and publishing the target position or velocity in separate topics since those will be called regularly. Do you have a preference on how such things are grouped?

Thanks!

clalancette:
> I created a Stepper class in the phidgets_api and I am modeling it after your Motor class. You added a Motors class since there are some phidgets that control multiple dc motors from one VINT port. There do not seem to be any stepper phidgets that control multiple motors from one port. Should I go ahead and create a Steppers class as well even though it is not really necessary just so it better matches the way dc motors are handled? Is the idea with these drivers that there will be one node per VINT port? An alternative might be to use a vector of serial numbers and hub ports when creating the Steppers class so that multiple steppers are controlled by one node, but perhaps just creating separate nodes, one for each stepper, is cleaner.

Generally, I prefer to only implement things that are needed.  So if you look at the gyroscope class, for instance, it only supports a single gyroscope since there are no phidgets that support more than one gyroscope.

In this case, though, it looks like there is a Phidget that supports more than one stepper: https://www.phidgets.com/?tier=3&catid=23&pcid=20&prodid=1033 .   Because of that I'd prefer to match the Motor/Motors split here.

[...]

PR #56, opened 2019-10-01
Add support for phidgets steppers

peterpolidoro:
Closes issue #55

### DIFF
--- a/phidgets_api/include/phidgets_api/stepper.hpp
+++ b/phidgets_api/include/phidgets_api/stepper.hpp
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) 2019, Howard Hughes Medical Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the copyright holder nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef PHIDGETS_API_STEPPER_H
+#define PHIDGETS_API_STEPPER_H
+
+#include <functional>
+
+#include <libphidget22/phidget22.h>
+
+#include "phidgets_api/phidget22.hpp"
+
+namespace phidgets {
+
+class Stepper final
+{
+  public:
+    PHIDGET22_NO_COPY_NO_MOVE_NO_ASSIGN(Stepper)
+
+    explicit Stepper(int32_t serial_number, int hub_port,
+                     bool is_hub_port_device, int channel,
+                     std::function<void(int, double)> position_change_handler,
+                     std::function<void(int, double)> velocity_change_handler);
+
+    ~Stepper();
+
+    int32_t getSerialNumber() const noexcept;
+
+    double getAcceleration() const;
+    void setAcceleration(double acceleration) const;
+    double getMinAcceleration() const;
+    double getMaxAcceleration() const;
+    bool stepControlMode() const;
+    void setStepControlMode(bool step_control_mode) const;
+    double getCurrentLimit() const;
+    void setCurrentLimit(double current_limit) const;
+    double getMinCurrentLimit() const;
+    double getMaxCurrentLimit() const;
+    uint32_t getDataInterval() const;
+    void setDataInterval(uint32_t data_interval_ms) const;
+    uint32_t getMinDataInterval() const;
+    uint32_t getMaxDataInterval() const;
+    bool getEngaged() const;
+    void setEngaged(bool engaged) const;
+    void enableFailsafe(uint32_t failsafe_time_ms) const;
+    uint32_t getMinFailsafeTime() const;
+    uint32_t getMaxFailsafeTime() const;
+    double getHoldingCurrentLimit() const;
+    void setHoldingCurrentLimit(double holding_current_limit) const;
+    double getPosition() const;
+    double getMinPosition() const;
+    double getMaxPosition() const;
+    void addPositionOffset(double position_offset) const;
+    double getRescaleFactor() const;
+    void setRescaleFactor(double rescale_factor) const;
+    void resetFailsafe() const;
+    double getTargetPosition() const;
+    void setTargetPosition(double target_position) const;
+    double getVelocity() const;
+    double getVelocityLimit() const;
+    void setVelocityLimit(double velocity_limit) const;
+    double getMinVelocityLimit() const;
+    double getMaxVelocityLimit() const;
+
+    void positionChangeHandler(double position) const;
+    void velocityChangeHandler(double velocity) const;
+
+  private:
+    int32_t serial_number_;
+    int channel_;
+    std::function<void(int, double)> position_change_handler_;
+    std::function<void(int, double)> velocity_change_handler_;
+    PhidgetStepperHandle stepper_handle_;
+
+    static void PositionChangeHandler(PhidgetStepperHandle stepper_handle,
+                                      void *ctx, double position);
+    static void VelocityChangeHandler(PhidgetStepperHandle stepper_handle,
+                                      void *ctx, double velocity);
+};
+
+}  // namespace phidgets
+
+#endif  // PHIDGETS_API_STEPPER_H
--- a/phidgets_steppers/README.md
+++ b/phidgets_steppers/README.md
@@ -1,0 +1,48 @@
+Phidgets stepper ROS 2 driver
+=============================
+
+This is the ROS 2 driver for Phidgets stepper.  The various topics, services, and parameters that the node operates with are listed below.
+
+Topics
+------
+* `/joint_states` (`sensor_msgs/JointState`) - A joint state message containing the current state of all steppers.
+
+Services
+--------
+* `/set_enabledXX` (`std_srvs/SetBool`) - Set true to enable/engage and false to disable/disengage stepper.  Must be set true before use.  One service is created for each stepper attached.
+* `/set_target_positionXX` (`phidgets_msgs/SetFloat64`) - Set the stepper target position when in position control mode.  One service is created for each stepper attached.
+* `/set_velocity_limitXX` (`phidgets_msgs/SetFloat64`) - Set the stepper velocity limit.  Must be set before use.  One service is created for each stepper attached.
+* `/set_accelerationXX` (`phidgets_msgs/SetFloat64`) - Set the stepper acceleration.  One service is created for each stepper attached.
+* `/set_current_limitXX` (`phidgets_msgs/SetFloat64`) - Set the stepper current limit.  Must be set before use.  One service is created for each stepper attached.
+* `/set_holding_current_limitXX` (`phidgets_msgs/SetFloat64`) - Set the stepper holding current limit.  One service is created for each stepper attached.
+* `/set_positionXX` (`phidgets_msgs/SetFloat64`) - Set the current position and target position a new value.  For example, setting the position to zero after homing.  One service is created for each stepper attached.
+* `/get_settingsXX` (`phidgets_msgs/GetStepperSettings`) - Get the values of the stepper settings. One service is created for each stepper attached.
+* `/get_setting_rangesXX` (`phidgets_msgs/GetStepperSettingRanges`) - Get the values of the stepper setting ranges. One service is created for each stepper attached.
+
+Parameters
+----------
+* `serial` (int) - The serial number of the phidgets stepper to connect to.  If -1 (the default), connects to any stepper phidget that can be found.
+* `hub_port` (int) - The phidgets VINT hub port to connect to.  Only used if the stepper phidget is connected to a VINT hub.  Defaults to 0.
+* `data_interval_ms` (int) - The number of milliseconds between acquisitions of data on the device (allowed values are dependent on the device).  Defaults to 250 ms.
+* `publish_rate` (double) - How often the driver will publish data on the ROS topic.  If 0 (the default), it will publish every time there is an update from the device (so at the `data_interval_ms`).  If positive, it will publish the data at that rate regardless of the acquisition interval.
+* `watchdog_interval_ms` (int) - The number of milliseconds allowed to elapse between calls to a watchdog timer.  Used as a failsafe to help prevent hardware damage in case of program crash. Defaults to 1000 ms with a range between 500-10000 ms.
+* `frame_id` (string) - The header frame ID to use when publishing the message.  Defaults to `stepper`.
+* `joint_nameXX` (string) - Name of each joint published in joint_states. Defaults to `joint_nameXX`.
+* `rescale_factorXX` (double) - Applies a factor to the [user units] per step to all movement parameters to make the units in your application more intuitive. Setting a new target position with a rescale factor, the stepper will move ((Target Position - Current Position) / RescaleFactor) steps. Defaults to 1.0.
+* `position_control_modeXX` (bool) - Set true for position control mode (step mode) or false for velocity control mode (run mode). Defaults to true.
+
+Command Line Examples
+---------------------
+
+```bash
+ros2 launch phidgets_steppers steppers-launch.py
+ros2 topic echo /joint_states
+ros2 service call /get_setting_ranges00 phidgets_msgs/GetStepperSettingRanges
+ros2 service call /set_current_limit00 phidgets_msgs/SetFloat64 "data: 1.0"
+ros2 service call /set_holding_current_limit00 phidgets_msgs/SetFloat64 "data: 0.5"
+ros2 service call /get_settings00 phidgets_msgs/GetStepperSettings
+ros2 service call /set_enabled00 std_srvs/SetBool "data: true"
+ros2 service call /set_velocity_limit00 phidgets_msgs/SetFloat64 "data: 5000"
+ros2 service call /set_target_position00 phidgets_msgs/SetFloat64 "data: 10000"
+ros2 service call /set_position00 phidgets_msgs/SetFloat64 "data: 0"
+```
--- a/phidgets_steppers/src/steppers_ros_i.cpp
+++ b/phidgets_steppers/src/steppers_ros_i.cpp
@@ -1,0 +1,474 @@
+/*
+ * Copyright (c) 2019, Howard Hughes Medical Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the copyright holder nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <chrono>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <stdexcept>
+
+#include <rclcpp/rclcpp.hpp>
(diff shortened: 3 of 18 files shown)


phidgets_api/include/phidgets_api/stepper.hpp
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (c) 2019, Open Source Robotics Foundation
```
clalancette: Minor note, here and elsewhere.  Since you've developed this code, you can change the copyright holder to be you (or whoever you work for).  It doesn't change the license or anything like that, it just gives a better idea of who contributed it and claims original authorship of it.

phidgets_steppers/src/steppers_ros_i.cpp
```diff
@@ -0,0 +1,381 @@
...
+}
+
+void StepperSetter::setEnabledCallback(const std_msgs::msg::Bool::SharedPtr msg)
+{
+    try
+    {
+        steppers_->setEngaged(channel_, msg->data);
+    } catch (const phidgets::Phidget22Error& err)
+    {
+        // If the data was wrong, the lower layers will throw an exception; just
```
clalancette: This is kind of the problem with setting things via topics; you can't easily tell if they succeeded or failed.  I think this would be better done with a service, which *does* give you a result, and thus you can tell if it succeeded.

I think the same is true for the rest of the `set` topics below.

phidgets_steppers/src/steppers_ros_i.cpp
```diff
@@ -0,0 +1,381 @@
...
+    {
+        steppers_->setHoldingCurrentLimit(channel_, msg->data);
+    } catch (const phidgets::Phidget22Error& err)
+    {
+        // If the data was wrong, the lower layers will throw an exception; just
+        // catch and ignore here so we don't crash the node.
+    }
+}
+
+void StepperSetter::addPositionOffsetCallback(
```
clalancette: This one seems like it could be a parameter instead; I don't see it as something you'll be changing on the fly very often.  If it *is* something you want to change often, then again this is probably better done as a service so you can get a pass/fail result.

phidgets_steppers/src/steppers_ros_i.cpp
```diff
@@ -0,0 +1,381 @@
...
+    {
+        steppers_->addPositionOffset(channel_, msg->data);
+    } catch (const phidgets::Phidget22Error& err)
+    {
+        // If the data was wrong, the lower layers will throw an exception; just
+        // catch and ignore here so we don't crash the node.
+    }
+}
+
+void StepperSetter::zeroPositionCallback(
```
clalancette: And finally, this one also seems like it would be better done as a service call, since you can tell if it fails.

phidgets_steppers/README.md
```diff
@@ -0,0 +1,37 @@
+Phidgets stepper ROS 2 driver
+=============================
+
+This is the ROS 2 driver for Phidgets stepper.  The various topics, services, and parameters that the node operates with are listed below.
+
+Topics
+------
+* `/set_enabledXX` (`std_msgs/Bool`) - Set true to enable/engage and false to disable/disengage stepper.  One topic is created for each stepper attached.
```
clalancette: I have some questions about these, but see the comments in the `.cpp` code below for more details.

phidgets_steppers/src/steppers_ros_i.cpp
```diff
@@ -0,0 +1,381 @@
...
+        stepper_data_to_pub_[channel].last_velocity_val = velocity;
+
+        if (publish_rate_ <= 0.0)
+        {
+            publishLatestJointStates();
+        }
+    }
+}
+
+void SteppersRosI::stoppedCallback(int channel)
```
clalancette: So this callback isn't really do anything.  If we aren't going to update any data about it, or publish anything, I'd say just don't bother with this at all, and we can simplify the whole stack by not passing it around or registering for it.

phidgets_api/include/phidgets_api/stepper.hpp
```diff
@@ -0,0 +1,116 @@
...
+    double getRescaleFactor() const;
+    void setRescaleFactor(double rescale_factor) const;
+    void resetFailsafe() const;
+    double getTargetPosition() const;
+    void setTargetPosition(double target_position) const;
+    double getVelocity() const;
+    double getVelocityLimit() const;
+    void setVelocityLimit(double velocity_limit) const;
+    double getMinVelocityLimit() const;
+    double getMaxVelocityLimit() const;
```
clalancette: While having this whole list of things implemented is nice, I'm generally of the opinion that we should only implement the bits that we need.  This reduces maintenance burden in the future; we don't have to change code that isn't there.  In this case, I would say that we should only implement what we are either setting or getting from `steppers_ros_i.cpp`; anything else is left for future expansion.

clalancette (changes requested):
I'll start out by saying that this is quite a good job.  The code is well organized, easy to read, has documentation and launch files, and generally looks like it will work well.

I do have some comments inline about simplifying it and rethinking the approach to setting values.  But otherwise, this is looking pretty good.

phidgets_steppers/src/steppers_ros_i.cpp
peterpolidoro: Good point! I will switch these to services. Do you think it is better to return a boolean status or to return something like the set value? That might depend on how out-of-range values are handled. If a user tries to set a value that is outside the range, do you think it would be better to return an error or just coerce the value to be within the range and then return that value so the user knows what it is? I would also like to be able to ask what all of the stepper settings are. That could be a separate service or alternatively whenever one value is set, all of the settings could be returned every time.

phidgets_steppers/src/steppers_ros_i.cpp
peterpolidoro: I can imagine situations where someone might want to add a position offset, but yes it might not be that useful and it might not hurt to remove it. The problem with making it a parameter is that the zeroPosition service will overwrite any offset that is set at the beginning. I will change it to a service.

phidgets_steppers/src/steppers_ros_i.cpp
clalancette: I think a boolean return is fine.  Also, I much prefer not to coerce values; instead, just try to do what the user asked and fail if it is out-of-range.  This could be made somewhat nicer by providing a way for the user to easily query what the range is, but it's probably not that important so I'd leave it out for now.

phidgets_steppers/src/steppers_ros_i.cpp
peterpolidoro: Yes I was trying to think of something useful I could do with a stopped callback and just left it there in case anyone had any ideas. At first I used it to publish 0.0 velocity on the JointState topic, but the changePosition callback already does that so that was redundant. I will remove it unless someone can think of something useful to do when the stepper stops.

phidgets_api/include/phidgets_api/stepper.hpp
peterpolidoro: I was hoping to use the min and max values to either coerce set values to be within the acceptable range or to provide a meaningful error message.

clalancette: Yeah, I'm totally fine with leaving what we are using.  So whatever you use, obviously keep.  Otherwise, we should just remove the bits we aren't using as it is effectively dead code at that point.

peterpolidoro:
I have a more general question about the naming of topics and services. I am still trying to wrap my head around the proper way to use data centric topics versus message centric topics. I do not know if that is considered as important in ROS as it is in the DDS middleware, but I see the potential advantages of being data centric. Many stepper controllers have very message centric APIs, like a "moveTo" command to move to an absolute position and a "moveBy" command to move relative distances. Using "get" and "set" topic names may be more data centric, but perhaps the most data centric would be to just publish something like the target position directly, without turning topic names into verbs at all. I have not surveyed enough ROS topic names to know what is considered best practice. I am also not sure of the best way to send relative move commands with either a "set_target_position" or a "target_position" topic, but other nodes could always calculate absolute position moves based on the current position, but some sort of relative move command might be cleaner in some cases. Also, is it worth trying to make all phidget motor packages have the same topic names so DC motors can be easily swapped for stepper motors or vice versa or to match topic names from other motor controller packages so phidgets controllers and other brand controllers can be used interchangeably? Or is it better to just match the phidgets APIs and then some sort of abstraction nodes iron out the differences between different motors or controllers?

clalancette:
So, I'll start out by saying that I haven't thought about message-centric vs. data-centric much in the context of this package.  For my use case, I've only been using "output" phidgets (like the temperature sensor, spatial sensor, etc).  I mostly ported the motor driver to libphidget22/ROS 2 because it was already there and I didn't want to degrade functionality, but it isn't something I'm using presently.

I'm going to answer some of your questions, but not necessarily in the order you asked them.

> Also, is it worth trying to make all phidget motor packages have the same topic names so DC motors can be easily swapped for stepper motors or vice versa

I don't think this is particularly worthwhile.  Most applications can't easily swap between DC motors and stepper motors.  So I wouldn't worry too much about this.

> or to match topic names from other motor controller packages so phidgets controllers and other brand controllers can be used interchangeably?

In general, the idea of ROS(2) drivers is to abstract away the differences between hardware so the higher levels of the stack don't have to care what they are talking to.  So from that perspective, yes, it would be worthwhile to look around the rest of the ROS/ROS 2 ecosystem and see what other motor drivers are doing, and match that if at all possible.

> perhaps the most data centric would be to just publish something like the target position directly

Agreed.  When I've written other ROS 2 nodes that talk to diff-drive motors, they take in a geometry_msgs/Twist command, and then the node is responsible for generating the left and right motor duty cycles to achieve that twist.  That doesn't *quite* work here, as you can't assume that the motor is being used as a diff-drive, but something more data centric like a target position does make sense to me.  But again, it's worth looking around the ROS ecosystem to see what the standards are.  (I'll note that the current motor driver does not do this, and instead just has a topic to set the motor_duty_cycle.  This may also want to be switched to a service.)

@mintar Any thoughts here?

peterpolidoro:
As far as naming goes, yes maybe it makes more sense to use "set" in service names rather than message topic names. So something like the name "joint_states" might be more appropriate for a topic name and anything like "set_current_limit" is more appropriate as a service name where you expect a response. So we could either use something like "target_position" to change the target position directly as a topic or "set_target_position" to change it as a service when you want to know if it succeeded. Then I guess it is debatable about which settings should be topics and which should be services, since most could go either way. It might be nice for calling nodes to know if certain values failed, but topics may be more convenient than services when you want something like a joystick node to publish velocity messages without worrying about responses.

Yes maybe it is a good idea to look at other ROS motor driver packages and match them. Which ones do you suggest? Perhaps something like the JointControllers? The Hardware Robot Information Model  [HRIM](https://github.com/AcutronicRobotics/HRIMl) is working on similar problems.

On a related note, do you think it makes sense to add more std_srvs to common_interfaces to use for setting values and getting a response? For example, it has a SetBool.srv, should we add SetFloat64.srv, etc?

peterpolidoro:
I see that the ros 1 std_msgs package has this warning in the documentation. Seems like a tradeoff between using messages that already exist versus creating new ones just so the field names are more meaningful. Would it be better to switch the phidgets topic messages to something more custom and descriptive?

> std_msgs contains wrappers for ROS primitive types, which are documented in the msg specification. It also contains the Empty type, which is useful for sending an empty signal. However, these types do not convey semantic meaning about their contents: every message simply has a field called "data". Therefore, while the messages in this package can be useful for quick prototyping, they are NOT intended for "long-term" usage. For ease of documentation and collaboration, we recommend that existing messages be used, or new messages created, that provide meaningful field name(s).

phidgets_steppers/README.md
```diff
@@ -0,0 +1,48 @@
+Phidgets stepper ROS 2 driver
+=============================
+
+This is the ROS 2 driver for Phidgets stepper.  The various topics, services, and parameters that the node operates with are listed below.
+
+Topics
```
mintar, line 6: Maybe worth renaming this to "Publishers".

mintar:
> @mintar Any thoughts here?

Regarding the topic types, I don't have any strong feelings one way or another (using std_msgs vs. defining a custom message with a single Float64 field). I think using `std_msgs/Float64` is fine - I believe that's what ros_control does, too. And clearly, a Twist message doesn't fit here.

Regarding topics vs. services: First off, a big disclaimer that I don't know how ROS2 handles services performance-wise. In ROS1, the best practice is IMO to use services for things where

* guaranteed message delivery is important,
* you need a result that's associated with the request you've sent,
* you only call the service once in a while (i.e., much less than 1 Hz), and
* the service returns quickly and doesn't need to be preempted (otherwise use an action).

On the other hand, use a topic if you regularly publish messages and if you don't care if a message gets dropped sometimes.

A good example of a service would be changing an operation mode, and a good example of a topic is publishing Twist messages to the `/cmd_vel` topic for commanding the motors on a mobile platform.

Is the user expected to publish messages to (for example) `set_target_position` at, say, 10 Hz or so? If yes, in ROS1 I'd definitely go with a topic then. But as I said, I don't know what the best practice is in ROS2.

clalancette:
> Is the user expected to publish messages to (for example) `set_target_position` at, say, 10 Hz or so? If yes, in ROS1 I'd definitely go with a topic then. But as I said, I don't know what the best practice is in ROS2.

I think your experience from ROS 1 translates over fairly well to ROS 2.  I think services are a little more performant in ROS 2, but if you would expect these controls to be a topic in ROS 1, then that makes it seem like they should be a topic in ROS 2 as well.

Before I cycle back on my recommendation, though; how do you handle the underlying call failing with a topic (in ROS 1)?  Is there some kind of feedback mechanism we could use?

peterpolidoro:
> Regarding the topic types, I don't have any strong feelings one way or another (using std_msgs vs. defining a custom message with a single Float64 field). I think using `std_msgs/Float64` is fine - I believe that's what ros_control does, too. And clearly, a Twist message doesn't fit here.

It seems like others have strong feelings about not using std_msgs or services:
https://github.com/ros2/common_interfaces/issues/73

phidgets_steppers/README.md
peterpolidoro: So instead of Topics you think we should have two headings, Publishers and Subscribers?

peterpolidoro:
I definitely want to be able to set the target position and velocity limit very quickly. So I should probably switch those back to being topics instead of services. If we are going to publish joint states then it might make sense to subscribe to a joint control message. So then phidgets motors would act like little joint controllers. Or I can keep position and velocity as independent topics if you prefer.

mintar:
> Before I cycle back on my recommendation, though; how do you handle the underlying call failing with a topic (in ROS 1)? Is there some kind of feedback mechanism we could use?

No, with topics there's no really good feedback mechanism. I think a good example of how this is handled in practice is the `cmd_vel` topic. The client (e.g., move_base) sends Twist messages and then monitors the robot's progress over a separate channel (for example, by watching the `odom` topic or the `odom -> base_link` tf transform) and checks if the robot hasn't moved before a timeout has expired.

To add to the confusion, let me bring in a third option apart from topics or services: actions. I actually like the idea of using an action for the stepper motors. The action would not return immediately, but instead only after the goal position is reached. Actions provide a feedback mechanism, are non-blocking and preemptible. For a client that just wants to move the stepper motor to a target position, an action would be perfect. For things like joystick teleoperation maybe a topic could be offered separately. The action type could be [FollowJointTrajectory](http://docs.ros.org/api/control_msgs/html/action/FollowJointTrajectory.html), but I don't know if the phidgets steppers actually support following a joint trajectory like that - I guess not, so I think defining a simpler custom action would be better.

About using JointStates for the control message: I also like that idea, however there is a caveat: According to the [JointState documentation](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html), the `position` and `velocity` arrays must either be empty or have the same length as the name array. If you want to support both setting position and velocity via the JointStates message, it becomes a bit unclear how to properly fill out the arrays. For example, let's say I want to set the `pan_joint` **position** to 0.7 and the `tilt_joint` **velocity** to 0.1. Do I fill out the fields like this?

```
name = ['pan_joint', 'tilt_joint']
position = [0.7, 0.0]
velocity = [0.0, 0.1]
effort = []
```

This would mean that 0.0 is a special value signifying "ignore me". We could of course use NaN instead, but either way is bad. We could also specify in the documentation that we expect either the position array or the velocity array to be filled, but not both. On the other hand, I believe at this point we're abusing the JointState message for something it wasn't intended to do, and it is probably much better to just create a `PositionCommand` and `VelocityCommand` message that only contains the fields we need (`name` and `position` / `name` and `velocity`).

Another design decision: Do we want to keep separate topics, each with the number of the motor in the topic name, or do we adress the joints by name, like in the JointState message (in this case, `pan_joint` and `tilt_joint`, or whatever else the user chooses)? The driver needs to know the names of the joints anyway in order to publish the JointState message (similar to what `phidgets_high_speed_encoder` already does). I think I favor using a single topic with joint names, but it's up to you to decide.

phidgets_steppers/README.md
mintar: Yes - it's pretty important to distinguish whether the node publishes a topic or subscribes to it. The ROS wiki documentation macros call the headings "Published Topics" and "Subscribed Topics", and I like that naming even better.

By the way, I just noticed the other phidgets READMEs also only call the heading "Topics". That should be changed too, I guess (but that's outside the scope of *this* PR).

peterpolidoro: Makes sense, I will change it. Do you have a link to a particular README that we can use as a template?

mintar: I think the rest of the README is fine, and it's consistent with the READMEs of the other packages here, so I would simply change the heading to "Published Topics" and be done with it. :)

peterpolidoro:
> To add to the confusion, let me bring in a third option apart from topics or services: actions. I actually like the idea of using an action for the stepper motors. The action would not return immediately, but instead only after the goal position is reached. Actions provide a feedback mechanism, are non-blocking and preemptible. For a client that just wants to move the stepper motor to a target position, an action would be perfect. 

Good idea! Yes, you are right, an action would be perfect for target position moves. I assume actions are fully implemented in ros 2, if so I will change it.  I will look at FollowJointTrajectory, maybe that would work well or be inspiration for a custom action. I think it would be ok to have to specify both a velocity and a target position for every target position move. It would be even more general, but perhaps overkill, to specify acceleration, velocity, and a target position for every move. It would be great to know when a move finishes, though, without having to poll the joint state velocity and it would be great to be able to abort a move.

In velocity control mode, however, it might be nice to just subscribe to a velocity_limit or cmd_vel topic to get super fast responses and the joint_states velocity could be used for feedback. It might be slightly confusing to have topics and actions that only work in one mode and not another, but perhaps making that clear in the documentation and using good names could help with that confusion. Or maybe using that topic or action could switch the mode automatically. If something publishes on the velocity topic the stepper could switch to velocity control mode and if something uses the action it could switch to position control mode. Which approach do you think would be more useful and less confusing?

Good point about using the joint names to address the motors. I also like the idea of using a single topic with joint names to address each of the motors. Using a message field seems a little cleaner and easier than using numbers in the topic names. Most of the phidget stepper controllers are only able to control a single motor, though. So this would only affect the more rare case of using the phidget stepper controller that can control multiple motors. For most cases, using multiple stepper motors will require using multiple stepper controllers. With the way these phidgets drivers are set up now, using multiple controllers will require multiple nodes, each with their own topics. It might be nice to figure out how to talk to all stepper controllers from a single node with a single topic, same with digital inputs, etc. Perhaps it is more fitting with the phidgets style, though, to just use one node per phidget. It still might be nice to use joint names to address the different stepper topics somehow instead of arbitrarily assigned numbers in the topic names.

phidgets_steppers/README.md
clalancette: > I think the rest of the README is fine, and it's consistent with the READMEs of the other packages here, so I would simply change the heading to "Published Topics" and be done with it. :)

Oh yeah, I totally agree with @mintar that this should be "Published Topics" and "Subscribed Topics".  I just forgot to do that when I did the original documentation.  I'll open a PR soon here to change the rest of the README's in this repository to have those titles.

clalancette:
> About using JointStates for the control message: I also like that idea, however there is a caveat: According to the [JointState documentation](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/JointState.html), the `position` and `velocity` arrays must either be empty or have the same length as the name array. If you want to support both setting position and velocity via the JointStates message, it becomes a bit unclear how to properly fill out the arrays. For example, let's say I want to set the `pan_joint` **position** to 0.7 and the `tilt_joint` **velocity** to 0.1. Do I fill out the fields like this?

Yeah, I agree that it is unclear how you would do that.  But I also think that we shouldn't use 0.0 as a special value; it may indeed be the case that you want to set the position or velocity to 0.0.  I'd argue that if you want to do something like you suggest, you have to send two separate messages with two different joints.  It's a little inefficient, but I also think it is something that probably isn't done that often.



> Another design decision: Do we want to keep separate topics, each with the number of the motor in the topic name, or do we adress the joints by name, like in the JointState message (in this case, `pan_joint` and `tilt_joint`, or whatever else the user chooses)? The driver needs to know the names of the joints anyway in order to publish the JointState message (similar to what `phidgets_high_speed_encoder` already does). I think I favor using a single topic with joint names, but it's up to you to decide.

So, I think the most "ROS" way to do it would be a single message and using the joint names; that matches what I've seen elsewhere.  However, the current way that the drivers in this repository do it seems to be one topic per device.  So it comes down to whether we want to be consistent amongst the drivers here, or consistent with how other drivers do it.  I actually don't have a strong opinion, so I can do either way here.

mintar:
@peterpolidoro wrote:

> In velocity control mode, however, it might be nice to just subscribe to a velocity_limit or cmd_vel topic

Small note: Don't call the new topic `cmd_vel` - that's a name that's customary for mobile bases, and is expected to have the topic type Twist.

I'm also a bit confused by the name `velocity_limit`. Is it actually a limit that controls the maximum velocity, or does it set the target velocity?

> If something publishes on the velocity topic the stepper could switch to velocity control mode and if something uses the action it could switch to position control mode. Which approach do you think would be more useful and less confusing?

Is there any reason why the user would want to switch the mode manually? If not, it could be something that we hide from the user and perform automatically.

mintar:
@clalancette wrote:

> Yeah, I agree that it is unclear how you would do that. But I also think that we shouldn't use 0.0 as a special value; it may indeed be the case that you want to set the position or velocity to 0.0. I'd argue that if you want to do something like you suggest, you have to send two separate messages with two different joints. It's a little inefficient, but I also think it is something that probably isn't done that often.

Yes, that was what I was trying to say. Put differently: we probably shouldn't use JointStates directly as a command topic type because it allows sending those confusing "mixed" commands. Instead, we should define custom messages that make it clear which fields can be set.

> So, I think the most "ROS" way to do it would be a single message and using the joint names; that matches what I've seen elsewhere. However, the current way that the drivers in this repository do it seems to be one topic per device. So it comes down to whether we want to be consistent amongst the drivers here, or consistent with how other drivers do it. I actually don't have a strong opinion, so I can do either way here.

I don't have a strong opinion either (just a slight preference for doing things the "ROS" way over consistency within our repo), so I guess it's up to @peterpolidoro to decide.

peterpolidoro:
> Small note: Don't call the new topic `cmd_vel` - that's a name that's customary for mobile bases, and is expected to have the topic type Twist.
> 
> I'm also a bit confused by the name `velocity_limit`. Is it actually a limit that controls the maximum velocity, or does it set the target velocity?

The Phidgets API uses the name 'velocity limit'. I think it is both a limit and a target depending on the control mode.

When the controller is in position control mode the velocity limit controls the maximum velocity and only positive velocity limit values are allowed. The direction of motion is set by the target position. For short moves, the motor may not accelerate all the way up to this value, but it tries.

When the controller is in velocity control mode the velocity limit sets a target velocity. The sign of the limit determines the direction of motion.

I could rename this topic to 'target_velocity' and only use it in velocity control mode.

> Is there any reason why the user would want to switch the mode manually? If not, it could be something that we hide from the user and perform automatically.

Most of the time, people will probably want to set the control mode once at the beginning and never change it. It could be useful to switch modes, though. For example, a user might want to drive the motors around in velocity control mode with a joystick and then once the motor is in position, then switch to position mode and make target position moves from software.

Phidgets tries to protect the motors by only allowing one network channel to attach to the controller so it cannot receive commands from multiple sources. Right now nothing prevents multiple ROS nodes from sending commands to the stepper node, defeating that safe guard. Allowing commands from multiple nodes seems more useful, but less safe. Switching the control mode automatically seems similar in that it would be more user friendly, but perhaps less safe. Do you think we should try to implement similar safe guards or just leave it less restricted and maybe more user friendly?

peterpolidoro:
> I don't have a strong opinion either (just a slight preference for doing things the "ROS" way over consistency within our repo), so I guess it's up to @peterpolidoro to decide.

I think we should consider changing all of the packages in the repo so only one node gets created for each Phidgets type. That could greatly reduce the number of nodes for some setups and look cleaner. So for example, there could be one DigitalInput node that reads from all the digital inputs, rather than having one separate node for each input. We could have one Stepper node to talk to all the stepper controllers rather than one separate node per stepper controller. Then we could use names or numbers as fields in the messages to address the channels rather than separate topic names. To implement this, we would need to do something like pass a vector of phidget serial numbers and VINT port numbers to use when creating the nodes instead of just one serial number and one port number per node.

mintar:
> The Phidgets API uses the name 'velocity limit'. I think it is both a limit and a target depending on the control mode.
> [...]
> I could rename this topic to 'target_velocity' and only use it in velocity control mode.

Thanks for clarifying. What about this?
1. a topic `target_position` that has fields both for the target position and the velocity limit
2. a topic `target_velocity` that has a field for target velocity

Same for actions.

> Most of the time, people will probably want to set the control mode once at the beginning and never change it. It could be useful to switch modes, though. For example, a user might want to drive the motors around in velocity control mode with a joystick and then once the motor is in position, then switch to position mode and make target position moves from software.

But in this use case, the user could just stop publishing to `target_velocity` and start publishing to `target_position` (or use the action), and the node could automatically switch the mode.

> Right now nothing prevents multiple ROS nodes from sending commands to the stepper node, defeating that safe guard.

With topics, there is indeed no way to prevent two nodes sending conflicting commands to the phidgets node, but I'd argue that it's the responsibility of the user to configure their node network correctly. It's the same with my favorite example, the mobile base `cmd_vel` topic. What people usually do is use a [topic_tools/mux](http://wiki.ros.org/topic_tools/mux) to switch between joystick input and move_base input.

With actions, it's easy: Whenever you get a new action goal, you abort the previous one (thereby notifying the client) and start the new one.

mintar:
Hi @peterpolidoro, did you close this PR intentionally? I think you've done a lot of valuable work here, and we should get it merged soon!

PR #57, opened 2019-10-01
Reformat using clang

peterpolidoro:


clalancette:
I was wondering why CI never showed this, and then I realized I never updated clang-check-style.py to look at `.hpp` files.  I'm going to update this, then open another PR that does exactly that.

Thanks for the contribution!

PR #60, opened 2019-10-18
Allow multiple hub ports per node

peterpolidoro:
What do you think of using a ChannelAddress struct when opening channels rather than individual parameters?

I realize this pull request breaks every package, but it would add flexibility for potentially adding more ChannelAddress members in the future and for perhaps using a vector of ChannelAddress when creating nodes. Right now, nodes are limited to a single port. If we pass a vector of ChannelAddress, then a node could access a whole set of ports.

I would need to go back and modify every package before you could accept this pull request. I am just throwing this change out for discussion.

### DIFF
--- a/phidgets_api/include/phidgets_api/accelerometer.hpp
+++ b/phidgets_api/include/phidgets_api/accelerometer.hpp
@@ -38,7 +38,7 @@
 
 namespace phidgets {
 
-class Accelerometer final
+class Accelerometer final : PhidgetChannel
 {
   public:
     PHIDGET22_NO_COPY_NO_MOVE_NO_ASSIGN(Accelerometer)
@@ -49,8 +49,6 @@ class Accelerometer final
 
     ~Accelerometer();
 
-    int32_t getSerialNumber() const noexcept;
-
     void getAcceleration(double &x, double &y, double &z,
                          double &timestamp) const;
 
@@ -59,7 +57,6 @@ class Accelerometer final
     void dataHandler(const double acceleration[3], double timestamp) const;
 
   private:
-    ChannelAddress channel_address_;
     std::function<void(const double[3], double)> data_handler_;
     PhidgetAccelerometerHandle accel_handle_;
 
--- a/phidgets_api/include/phidgets_api/accelerometer.hpp
+++ b/phidgets_api/include/phidgets_api/accelerometer.hpp
@@ -44,7 +44,7 @@ class Accelerometer final
     PHIDGET22_NO_COPY_NO_MOVE_NO_ASSIGN(Accelerometer)
 
     explicit Accelerometer(
-        int32_t serial_number, int hub_port, bool is_hub_port_device,
+        const ChannelAddress &channel_address,
         std::function<void(const double[3], double)> data_handler);
 
     ~Accelerometer();
@@ -59,7 +59,7 @@ class Accelerometer final
     void dataHandler(const double acceleration[3], double timestamp) const;
 
   private:
-    int32_t serial_number_;
+    ChannelAddress channel_address_;
     std::function<void(const double[3], double)> data_handler_;
     PhidgetAccelerometerHandle accel_handle_;
 
--- a/phidgets_api/include/phidgets_api/analog_input.hpp
+++ b/phidgets_api/include/phidgets_api/analog_input.hpp
@@ -43,8 +43,7 @@ class AnalogInput final
   public:
     PHIDGET22_NO_COPY_NO_MOVE_NO_ASSIGN(AnalogInput)
 
-    explicit AnalogInput(int32_t serial_number, int hub_port,
-                         bool is_hub_port_device, int channel,
+    explicit AnalogInput(const ChannelAddress &channel_address,
                          std::function<void(int, double)> input_handler);
 
     ~AnalogInput();
@@ -58,8 +57,7 @@ class AnalogInput final
     void voltageChangeHandler(double sensorValue) const;
 
   private:
-    int32_t serial_number_;
-    int channel_;
+    ChannelAddress channel_address_;
     std::function<void(int, double)> input_handler_;
     PhidgetVoltageInputHandle ai_handle_;
 
--- a/phidgets_api/include/phidgets_api/analog_input.hpp
+++ b/phidgets_api/include/phidgets_api/analog_input.hpp
@@ -38,7 +38,7 @@
 
 namespace phidgets {
 
-class AnalogInput final
+class AnalogInput final : PhidgetChannel
 {
   public:
     PHIDGET22_NO_COPY_NO_MOVE_NO_ASSIGN(AnalogInput)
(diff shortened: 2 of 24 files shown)


peterpolidoro:
As I read more about node parameters, this change may not make sense.

I was hoping to be able to start a node with a vector of Phidget channel addresses, rather than just a single address. So then, for example, we could have one node of all DigitalInputs, rather than one node for each DigitalInput. Although perhaps using lots of nodes instead is not a big deal.

If we do want to be able to open several channels on different ports per node, then we need some way of addressing them all. Using a single vector of structs as a class constructor parameter is cleaner than using several parameters that are vectors of simple types.

When launching a node, however, we can only pass in node parameters that are vectors of simple types. So we would need a node parameter that is a vector of serial numbers, a node parameter that is a vector of hub port numbers, etc. The parameter vectors would need to be equal length and each channel address would be the parameter values at a particular index, for example channel_address[2] = (serial[2], hub_port[2], ...). If the parameters to the nodes have to be a set of vectors, then maybe the parameters to the class constructors should just be a set of vectors too.

peterpolidoro:
Actually, what do you think about making another big change and not use parameters to specify which channel or channels to open when launching a node? Instead, we could just launch a node without opening any channels and then later use an action to specify which channels to attach.

Each channel could be specified by a ChannelAddress message, containing the serial number, hub port, etc. Then you could send an array of ChannelAddresses to each node as an action goal. Part of the ChannelAddress message could be a user specified name for each channel. Then future messages to that node could use that name when specifying a channel.

I will start to implement this in this pull request just to help explain what I am suggesting.

mintar:
I haven't looked into this part of the ROS2 code deeply enough to understand what this PR is trying to achieve, so I'll leave that up to @clalancette. But I'll add some specific comments:

> I realize this pull request breaks every package

Just a note on that, without commenting on whether this PR is a good idea: The driver has not been released into Dashing yet, so I guess now is still a good time to do breaking changes (before the package has been widely adopted).

> Actually, what do you think about making another big change and not use parameters to specify which channel or channels to open when launching a node? Instead, we could just launch a node without opening any channels and then later use an action to specify which channels to attach.

I'm very skeptical of this approach. In my opinion, things like the serial number should be configured statically using parameters. Having an action do this would 

Also, if I understand you correctly, you propose using actions to work around the limitation that parameters must be simple types. Wouldn't that just shift the problem? How would you configure the action client node, so that it knows which channels to open? If you use parameters, you got the same problem as before.

If I understood you correctly, and your only problem is that parameters must be simple data types, there's a solution to that (at least on ROS1, but maybe this works on ROS2 as well?): You can get a whole subtree of parameters as a dictionary/map. So you could structure your parameters like this:

```yaml
phidgets_node:
  - some_other_parameter: 42
  - channels:
    - my_channel_name1:
      - serial_number: 123456789
      - int hub_port: 23
      - is_hub_port_device: true
      - channel: 7
    - my_channel_name1:
      - ...
```

The node could then just get the `phidgets_node/channels` parameter and iterate over the resulting map.

peterpolidoro:
> I haven't looked into this part of the ROS2 code deeply enough to understand what this PR is trying to achieve

The goal of this pull request is to remove the limitation of only one hub port per node. I changed the name of the pull request to make that more clear. I would like, for example, to be able to open multiple motors, each connected to its own hub port, in a single node, then address each motor by joint name in a message field.

I would like to specify each hub port in the node using a channel address struct. Then we can simply use a vector of these channel address structs to group a set of channels for the node, even though each channel is using a separate hub port.

peterpolidoro:
> If I understood you correctly, and your only problem is that parameters must be simple data types, there's a solution to that (at least on ROS1, but maybe this works on ROS2 as well?)

Yes that is correct, I think there is an issue in ROS2 of only allowing parameters with simple data types or homogenous arrays of simple data types. Perhaps I am misunderstanding this page however: [ROS2 Parameters](https://design.ros2.org/articles/ros_parameters.html)

If that limitation is true, then sending a message to a node seems like a much cleaner way of communicating complex data structures to that node rather than using lots of vector parameters of simple types.

> I'm very skeptical of this approach. In my opinion, things like the serial number should be configured statically using parameters.

Why so skeptical? Phidget nodes can be opened and closed dynamically at run time. Why insist on only allowing them to configured statically when a node is launched? For example, it might be nice to plug in a new Phidget and tell a running node to open it and add it to its list of channels that it can control. Statically configuring them will cause a node to fail to launch if a Phidget has not been plugged in yet.

You are right that there is still the problem of getting the information into the configuration node so it can send a message to the Phidgets node. Perhaps the configuration node can read a yaml file to act more like ROS1 parameters or get the information from the user in a GUI or something.

clalancette:
So, there's a lot of discussion above.  I'd like to make a few points:

1.  I'm not against using a `ChannelAddress` structure, as that would clean up the call signatures a bit.
1.  As it stands, whether you need one or many nodes per `DigitalInput` (for instance) depends on how they are connected.  If they are connected via an IK, for instance, then as it stands the code will detect all of them behind the IK and deal with them all.  If you want to use all of the hub ports on a VINT hub as digital input, then yeah, you need one node per-port.
1.  As far as parameters in ROS 2 goes, there are ways to get somewhat more complex things.  In particular, you can get `maps` so that anything below a certain heading comes back.  For instance, you can have:

```
foo:
   x: 4
   y: 5
```

And then you can get a map of `foo.x` and `foo.y`.  There's an example on how to do something like this at https://github.com/ros2/teleop_twist_joy/blob/dashing/src/teleop_twist_joy.cpp#L85.

1.  I'm with @mintar in that I would much rather set everything up using static parameters rather than trying to do something dynamic.  While being dynamic is nice, it adds a lot of complexity and it isn't clear to me that this complexity is worthwhile.  And I usually want to keep things as simple as possible until a piece of functionality is really needed.  @peterpolidoro does your current application require dynamic attachment of devices?

I'm sure I've missed some of the points you've discussed above, so feel free to re-ping me on anything else you'd like me to comment on.

peterpolidoro:
> 3. As far as parameters in ROS 2 goes, there are ways to get somewhat more complex things.  In particular, you can get `maps` so that anything below a certain heading comes back.

Oh cool, I will check out those parameter maps, thanks! I am fine with using static parameters to open devices during launch, I was mostly worried about the limitations of parameters with only simple types. I am building some systems where the users may want to be able to connect and disconnect Phidgets during run time. I agree with both of you, though, that it is better to keep things simple, then add functionality when needed. We can always add an option of dynamic reattachment later if some applications need it.

I am still making changes to this branch to make nodes work with multiple ports. You may find the changes too significant and if so, that is fine, I can abandon this approach. My application requires many stepper motors and digital inputs each connected to its own port, so it seems really nice to me to only have a couple of nodes rather than lots of nodes, but perhaps with the way nodes work in ROS2 using lots of nodes would work just fine. I will keep making changes until I can get everything to compile and run again and then you tell me if you think all of the changes are worth it or not. Thanks for all of your feedback!

PR #61, opened 2019-10-18
Add developing documentation to README

peterpolidoro:
I am not sure this documentation is appropriate to add to the README. If not, feel free to reject this pull request. I find it useful cat the README to remind myself of the reformatting commands before committing. Perhaps this documentation could go somewhere else or should just be remove entirely.


clalancette (approved):
Seems reasonable to me.  I'd like to change `clang-check-style.py` to take a flag so that the `find` command isn't necessary, but we can do that in a separate PR.

clalancette:
@mintar Any thoughts?  Otherwise I'll merge this later today.

mintar:
LGTM!

PR #63, opened 2019-10-24
Create PhidgetChannel and PhidgetChannels base classes

peterpolidoro:
@clalancette How do you feel about using some base classes to simply some of the api files? I removed all of the packages except motor in this pull request to make it easier to see the effects of using the base classes.

The PhidgetChannel base class allowed me to move some of the methods that were duplicated across files (such as getSerialNumber) into a single location.

The PhidgetChannels base class allowed me to remove the plural api files (such as motors.hpp and motors.cpp).

I have tested these changes and they compile and run as they did before the changes (except I temporarily removed the functionality of opening all of the channels on a multiple channel device, but I would add that back later).

Do you find these changes too extreme or do you approve of the simplifications? If you approve then I can apply them to the rest of the packages in my other pull requests.


clalancette:
So overall, I'm not a huge fan of making a base class that all of the classes in `phidgets_api` are descendant from.  The main reason is that there are more differences between the classes then there are similarities, so it just doesn't make sense to me to tie them all together.

You could make the argument that the duplicated code for `getDeviceSerialNumber` littered everywhere is too much duplication, and I would (mostly) agree with you.  But I think that the right solution for that is another helper function in `phidget22.hpp` that calls the serial number API and throws an exception if it fails.

That also goes for the addition of the templated `PhidgetsChannels` class.  While that would again reduce some of the duplication amongst the `Motors`, `DigitalInputs`, etc. classes, I think it does so at the expense of readability.  Again, with the exception of the `getFooCount()` APIs, those classes are more different than similar.  They are also carefully designed to make sure that RAII cleans up even in the failure cases.

I'm still on board with adding in the `PhidgetsChannel` class to consolidate all of the information together.  But I'd appreciate a PR that did just that, so it is easy to see just those changes.

Thanks for your efforts here so far.  I'm going to close this PR out, but feel free to keep commenting or reopen if you want to discuss this more.

peterpolidoro:
> So overall, I'm not a huge fan of making a base class that all of the classes in `phidgets_api` are descendant from. The main reason is that there are more differences between the classes then there are similarities, so it just doesn't make sense to me to tie them all together.

The one thing that ties them all together is that they all inherit functions from the Phidget class in the API from Phidgets and they all have a channel address. If you prefer using helper functions to inheritance, though, then you are right, there is no reason for a PhidgetChannel class. The two approaches are basically equivalent, so I am happy to go with whatever you prefer. The ROS 1 version of these drivers uses a Phidget base class that all other API classes inherit from, did that approach have some issues?

> That also goes for the addition of the templated `PhidgetsChannels` class. While that would again reduce some of the duplication amongst the `Motors`, `DigitalInputs`, etc. classes, I think it does so at the expense of readability. Again, with the exception of the `getFooCount()` APIs, those classes are more different than similar. They are also carefully designed to make sure that RAII cleans up even in the failure cases.

The thing that ties the all of the plural classes together is that they all contain a vector of unique pointers and almost all of their methods are just wrappers around the matching method in the singular class, adding an index parameter. The duplication is pretty minimal in most of the singular classes, but gets pretty heavy in classes with lots of methods. I am not sure I understand what you mean about the plural classes being very different from each other and cleaning up in failure cases. Cleaning up properly does sound important, how does that work?

The PhidgetsChannels class does not really do much now, but I was hoping to add some functionality to it. For example, I thought it would be nice to add a name string to each channel address with the ability to sort the vector by the name. Then the name could be used to quickly fetch a member of the vector. So motors could be referenced by their joint names or somewhat arbitrary vector index.

I thought using a PhidgetsChannels class would make it easier to give all nodes the ability to interface to multiple devices of the same type. For example, one node could read from multiple temperature sensor Phidgets, without having create a Temperatures class and wrap all of the temperature methods. Do you prefer creating a Temperatures class instead? Or do you prefer that nodes just interface with a single Phidget, unless that Phidget contains multiple channels?

> I'm still on board with adding in the `PhidgetsChannel` class to consolidate all of the information together. But I'd appreciate a PR that did just that, so it is easy to see just those changes.

I am not sure I understand what you mean by this either. Do you mean not using helper functions and inherit instead?

Sorry to keep slamming you with so many changes and requests for feedback. I just need to get these drivers up and running well enough for me to continue with the next stages of my projects. Thanks for all of your help and advice.

I will be at ROSCON next week if you are there and are interested in talking about any of this in person.

PR #70, opened 2020-03-20
Composite Design Pattern Discussion

peterpolidoro:
Hi. I finally have some more time to work on these drivers. I am opening this pull request just to discuss some possible changes I could make.

For some of my projects I need to use the Dashing distribution and make some breaking changes. That may mean I need to just keep my own fork and keep changes separate from this repository, at least for now. I would like to get your input, though, in case you want any changes, perhaps on future distributions.

After lots of experimenting with real hardware, I found that the [composite pattern](https://en.wikipedia.org/wiki/Composite_pattern) seems to work well with Phidgets. That way a collection of heterogeneous Phidgets is treated in a very similar way as a single Phidget. For example, a Stepper Phidget can be extended to a Stepper Joint Phidget composite, that includes both a home switch DigitalInput and an optional limit switch DigitalInput. The composite can be open and closed and queried just as if it was a single Phidget.

There are a couple of ways to implement the composite pattern. One way is just to live in the C++ classes world and create a PhidgetComponent base class that a Phidget class and a PhidgetComposite class both inherit from. Then all the Phidget callbacks are handled within those classes.

Another way to implement the composite pattern might be to push everything to the ROS abstraction layer and form a composite from a collection of ROS nodes rather than C++ classes directly. That seems to be more of the original intent of the drivers in this repository. Composing with nodes might add flexibility and transparency as to what is happening within the composite. It may have some performance penalties, though, of having Phidget callbacks send ROS messages which then trigger ROS callbacks instead of just using the Phidget callbacks directly.

Phidgets breaks apart systems into smaller pieces than most other controllers and sensors. Many other stepper controllers, for example, have home switch and limit switch inputs that are handled internally in the controller. This makes it a little harder to observe what is happening within the controller, but it improves the response times to the digital inputs significantly.

So implementing the composite design pattern with C++ classes is probably more similar to how other ROS drivers would be written and might have better performance. Implementing the composite design pattern with ROS nodes might be more general and people using these drivers might need to know less about Phidgets and use more ROS tools.

Do you have a preference on which approach would be better or is there yet another way that would be preferred? Do you have interest in me making changes to the dashing branch of this repository or should I keep everything in my own fork and you can decide if you want changes later in a future distribution once I have worked out more details?

### DIFF
--- a/phidgets_api/include/phidgets_api/phidget22.hpp
+++ b/phidgets_api/include/phidgets_api/phidget22.hpp
@@ -55,7 +55,7 @@ class Phidget22Error final : public std::exception
     std::string msg_;
 };
 
-struct ChannelInfo
+struct PhidgetInfo
 {
   int32_t serial_number{ PHIDGET_SERIALNUMBER_ANY };
   int hub_port{ PHIDGET_HUBPORT_ANY };
@@ -65,13 +65,6 @@ struct ChannelInfo
 
 class PhidgetComponent
 {
-public:
-  PhidgetComponent(std::string name);
-  virtual bool hasHandle() const override;
-  virtual void setOnAttachHandler(std::function<void(PhidgetHandle, void *)> on_attach_handler) override;
-private:
-  std::string name_;
-  void on_attach_handler();
 };
 
 class Phidget : public PhidgetComponent
--- a/phidgets_api/include/phidgets_api/phidget22.hpp
+++ b/phidgets_api/include/phidgets_api/phidget22.hpp
@@ -32,6 +32,7 @@
 
 #include <stdexcept>
 #include <string>
+#include <functional>
 
 #include <libphidget22/phidget22.h>
 
@@ -52,6 +53,34 @@ class Phidget22Error final : public std::exception
 
   private:
     std::string msg_;
+};
+
+struct ChannelInfo
+{
+  int32_t serial_number{ PHIDGET_SERIALNUMBER_ANY };
+  int hub_port{ PHIDGET_HUBPORT_ANY };
+  bool is_hub_port_device{ false };
+  int channel{ PHIDGET_CHANNEL_ANY };
+};
+
+class PhidgetComponent
+{
+public:
+  PhidgetComponent(std::string name);
+  virtual bool hasHandle() const override;
+  virtual void setOnAttachHandler(std::function<void(PhidgetHandle, void *)> on_attach_handler) override;
+private:
+  std::string name_;
+  void on_attach_handler();
+};
+
+class Phidget : public PhidgetComponent
+{
+};
+
+class PhidgetComposite: public PhidgetComponent
+{
+
 };
 
 namespace helpers {


clalancette:
> Another way to implement the composite pattern might be to push everything to the ROS abstraction layer and form a composite from a collection of ROS nodes rather than C++ classes directly. That seems to be more of the original intent of the drivers in this repository. Composing with nodes might add flexibility and transparency as to what is happening within the composite. It may have some performance penalties, though, of having Phidget callbacks send ROS messages which then trigger ROS callbacks instead of just using the Phidget callbacks directly.

You are totally correct on this point; there is meant to be one ROS 2 node per Phidget device attached to the system.  For VINT Hub devices, you are then *forced* to compose them, since only one process can open the USB device at a time.  I've found this model to work fairly well so far in our application, though we are only using a handful of Phidgets at this time.

It might be helpful here to describe more concretely what you are trying to achieve.  Maybe we can come up with some solutions for you that fit into the existing structure.  I'm certainly willing to entertain notions of changing the class structure, but there would have to be good reasons to do so.

peterpolidoro:
Hi. I do like the idea of using a ROS node for every Phidget. It seems convenient to use ROS for both constructing a device from a collection of Phidgets and for connecting a set of devices into a larger system. Since you know more about the internal workings of ROS than I do, I was just curious to know what you think the performance penalties might be for pushing all of the low-level interactions to the ROS abstraction layer instead of just using C++ classes directly. Perhaps any ROS performance penalties are insignificant, though, when compared to the penalties of using Phidgets in the first place.

As an example, I would like to have a single ROS interface to a variety of XYZ-stage controllers. Perhaps more generally it would just be an instance of a multi-axis joint controller. An XYZ-stage could be constructed from three linear actuators. Each linear actuator could have one stepper motor and two limit switches. So the XYZ-stage would have three stepper motors and six limit switches total.

The company Trinamic makes a single little IC chip capable of controlling three stepper motors with six limit switches. Each limit switch is connected to a hardware interrupt in the IC that stops the motor with very little delay when a limit switch gets hit.

If I also want to use Phidgets as another alternative for controlling the XYZ-stage, I would need three Stepper Phidgets and six DigitalInput Phidgets. Phidgets are very convenient for quickly constructing custom controllers from off-the-shelf parts, but they have the disadvantage of having significant delays from passing lots of messages around over serial communication. A limit switch connected to a DigitalInput needs to pass a message from the Phidget hub over USB to the host computer, which then handles that message in software triggering a Phidget callback, which then needs to pass another message over USB back to the Phidget hub, which then sends another serial message to the Phidget Stepper controller telling it to stop the motor. That is a huge delay compared to a hardware interrupt in an IC chip.

If every Phidget is a ROS node, then the delay would be even bigger. After the DigitalInput Phidget callback, the ROS layer needs to serialize a ROS message and perhaps send it to a joint controller node which deserializes it and triggers a ROS callback, which then serializes and sends another message to the stepper node which deserializes it and triggers another ROS callback, which then sends the message over USB to the Phidget hub.

Perhaps ROS is efficient enough, especially when everything is in one process, that the extra ROS message passing is not very significant when compared to the Phidget message passing.

If you think the ROS delays could be significant, though, I was thinking an alternative might be to just assemble something like an XYZ-stage Phidget controller using C++ classes directly in a composite design pattern. So a linear actuator controller could just be a C++ class containing an instance of a Stepper class and two instances of DigitalInput classes. An XYZ-stage controller would just contain three instances of the linear actuator controller classes. Using the composite design pattern would allow you to call xyz_stage_controller.open() just as you would call digital_input.open(), recursively calling the open() method automatically on all of the phidgets and phidget composites contained in the xyz_stage.

I could then add a ROS interface to the XYZ-stage Phidget composite class, which would hopefully be almost identical to the ROS interface to something like an XYZ-stage triniamic C++ driver. 

If the ROS delays are not very significant, then perhaps making every Phidget a ROS node is the most flexible and convenient option. If they are significant, though, then perhaps it might be nice to just have a single ROS node interface to a larger collection of Phidgets and not use ROS inside the device to pass messages between Phidgets.

Do you think ROS delays are significant when compared to Phidget delays?

clalancette:
> Perhaps ROS is efficient enough, especially when everything is in one process, that the extra ROS message passing is not very significant when compared to the Phidget message passing.

It is definitely the case that nodes within the same process pass messages much more efficiently than in separate processes.  In point of fact, we could go further with the ROS 2 port and optimize it further for the multiple nodes in one process case (basically by turning on intra-process comms).  But since I haven't run into performance issues here, I haven't bothered.

> Do you think ROS delays are significant when compared to Phidget delays?

The real truth of the matter is that I don't know.  I don't know how significant the Phidget internal delays are, the Phidget->USB delay, the kernel USB->userland delay, or the ROS 2 -> ROS 2 delays are.  Someone would have to spend some time to setup a test that measures all of these delays and determines what the percentages are.

peterpolidoro:
Do you have any links to documentation describing how to turn on intra-process communication? Maybe that is something I can add. Also do you have any recommendations on profilers I could use to try to measure all of those delays? Thanks!

clalancette:
> Also do you have any recommendations on profilers I could use to try to measure all of those delays? Thanks!

To be clear, this is the thing to do first.  It's not worth doing any optimization until you know where the bottlenecks are.

As for your question about how to do it, the thing you really want to time is the time since the event happened at each step.  The most exact way (and the most work) is to have a GPIO that fires at each event, and then collect the timing of thos GPIOs.  But this requires a whole setup to do.

As far as profiling just the software side, you can use the `perf` tool on Linux to get some idea of CPU time.  http://www.brendangregg.com/FlameGraphs/cpuflamegraphs.html gives some examples on how to do that for a flame graph.  You are more interested in latency here, so maybe some of the stuff on http://www.brendangregg.com/perf.html will be helpful.

> Do you have any links to documentation describing how to turn on intra-process communication? Maybe that is something I can add.

https://index.ros.org/doc/ros2/Tutorials/Intra-Process-Communication/ gives some idea on how to enable it.

peterpolidoro:
Thanks I will check those out. 

If I can get my oscilloscope out of quarantine lockdown I will make some timing measurements with real hardware too.

PR #72, opened 2020-06-01
Add rosdep to the Dockerfile for Dashing.

clalancette:


mintar (approved):
LGTM!

clalancette:
Thanks, merging.  Note that I won't do a new release onto the buildfarm for this, since it only affects the source builds.

clalancette:
(also note that Fpr failed because newer gcc on Ubuntu Focal complains about code down in libphidget22; I'll fix that in another PR)

PR #74, opened 2020-06-02
Update libphidget22

clalancette:
While looking at the warnings that libphidget22 was producing on Ubuntu 20.04, I decided to try to update to the latest libphidget22.  That fixed some warnings, but introduced some others.  It still seemed worthwhile to me to keep up with the upstream, however, hence this PR.

This updates libphidget22 to the newest version available from https://www.phidgets.com/downloads/phidget22/libraries/linux/libphidget22/ .  It also patches that version to get rid of all warnings.  Finally, it does a small amount of cleanup in the header files that I noticed while doing this update.  Compiles without warning for me on Ubuntu 20.04, though I haven't been able to do any testing of it.


mintar (approved):
LGTM!

PR #78, opened 2020-06-08
Update Travis CI + Dockerfile to Foxy

mintar:
@clalancette: Feel free to squash this PR directly onto Foxy if you agree. Tested locally.

### DIFF
--- a/Dockerfile-foxy
+++ b/Dockerfile-foxy
@@ -1,16 +1,16 @@
-FROM ros:dashing-ros-core
+FROM ros:foxy-ros-core
 
 RUN apt-get update && apt-get install -y \
-    build-essential clang-format python3-colcon-common-extensions python3-rosdep
+    build-essential file clang-format python3-colcon-common-extensions python3-rosdep
 
 # Create ROS workspace
 COPY . /ws/src/phidgets_drivers
 WORKDIR /ws
 
 # Use rosdep to install all dependencies (including ROS itself)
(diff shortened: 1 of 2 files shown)


Dockerfile-foxy
```diff
@@ -1,16 +1,16 @@
-FROM ros:dashing-ros-core
+FROM ros:foxy-ros-core
 
 RUN apt-get update && apt-get install -y \
-    build-essential clang-format python3-colcon-common-extensions python3-rosdep
+    build-essential file clang-format python3-colcon-common-extensions python3-rosdep
```
clalancette, line 4: @mintar Question; did you actually need this change for things to succeed in Foxy?  Just wondering why.

mintar: Oh sorry, I completely missed your remark. Yes, this change is necessary because the `configure` script of `libphidget22` needs the `file` command. I assume it was installed by default in the `ros:dashing-ros-core` image, so there was no need to install it explicitly there.

The log output during the colcon build is:

```
Starting >>> libphidget22
--- stderr: libphidget22
/ws/build/libphidget22/libphidget22-src/configure: line 6324: /usr/bin/file: No such file or directory
---
```

It's actually more of a warning, since the build and tests also succeed when `file` is not installed. Just seemed cleaner to me to install it anyway.

clalancette: OK, no worries!  Just checking.  Thanks for the explanation.

clalancette (approved):

PR #79, opened 2020-06-20
Use private node handler for params.

edavies64:
I noticed that the high speed encoder library doesn't respect the parameters set in a launch file because they are supposed to private not public. This fix allows tick2rad and the serial number to be properly set for each channel. Also changed the printout for ROS_INFO about channel name to properly output desired information. 
```
[ INFO] [1592680369.498374641]: Channel 0 Name: 'back_left_wheel_joint', Tick2rad: '0.001088' 
```

This has been tested and works on our robot. 
Also, the phidgets wiki reference for parameters are outdated. They still reference 0.7.9 instead of new names in 0.7.10. In particular "serial_name" has been changed to "serial_number"

### DIFF
--- a/phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
+++ b/phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
@@ -24,7 +24,8 @@ HighSpeedEncoderRosI::HighSpeedEncoderRosI(ros::NodeHandle nh,
         sprintf(str, "joint%u_tick2rad", i);
         nh_private_.getParam(str, joint_tick2rad_[i]);
 
-        ROS_INFO("Channel %u Name: '%s', Tick2rad: '%f'", i, joint_names_[i].c_str(), joint_tick2rad_[i]);
+        ROS_INFO("Channel %u Name: '%s', Tick2rad: '%f'", i,
+                 joint_names_[i].c_str(), joint_tick2rad_[i]);
     }
 
     int serial_number = -1;
--- a/phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
+++ b/phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
@@ -19,18 +19,18 @@ HighSpeedEncoderRosI::HighSpeedEncoderRosI(ros::NodeHandle nh,
     {
         char str[100];
         sprintf(str, "joint%u_name", i);
-        nh_.getParam(str, joint_names_[i]);
+        nh_private_.getParam(str, joint_names_[i]);
 
         sprintf(str, "joint%u_tick2rad", i);
-        nh_.getParam(str, joint_tick2rad_[i]);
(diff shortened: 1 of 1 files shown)


seebq:
Hey @mintar I paired with @edavies64 on this fix and can confirm that without this, it doesn't load the correct params with a launch file.

Wonder why the CI didn't build though?

Will there be more renaming and changes to param names? if so, we may keep our fork.

phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
```diff
@@ -19,18 +19,18 @@ HighSpeedEncoderRosI::HighSpeedEncoderRosI(ros::NodeHandle nh,
...
         sprintf(str, "joint%u_name", i);
-        nh_.getParam(str, joint_names_[i]);
+        nh_private_.getParam(str, joint_names_[i]);
 
         sprintf(str, "joint%u_tick2rad", i);
-        nh_.getParam(str, joint_tick2rad_[i]);
+        nh_private_.getParam(str, joint_tick2rad_[i]);
 
-        ROS_INFO("Channel %u: '%s'='%s'", i, str, joint_names_[i].c_str());
+        ROS_INFO("Channel %u Name: '%s', Tick2rad: '%f'", i, joint_names_[i].c_str(), joint_tick2rad_[i]);
```
mintar:
```suggestion
        ROS_INFO("Channel %u Name: '%s', Tick2rad: '%f'", i,
                 joint_names_[i].c_str(), joint_tick2rad_[i]);
```

This change should fix [this code style test](https://travis-ci.org/github/ros-drivers/phidgets_drivers/jobs/700421992#L1312-L1316).

mintar (changes requested):
Thanks for your PR! Could you please fix the change I suggested and drop me a message, so I can squash + merge? I've also updated the wiki as you suggested; I don't believe `serial_name` was ever correct, it should always have been `serial_number`.

mintar:
@ros-pull-request-builder Retest this please

mintar:
> Will there be more renaming and changes to param names? if so, we may keep our fork.

No, apart from obvious bug fixes like this one there won't be any changes to parameter names. If I decide to change parameters (for example, renaming `PUBLISH_RATE` to `publish_rate`), I'll follow a tick-tock scheme, that is if the parameter `publish_rate` is not found, I'll look for the parameter `PUBLISH_RATE` instead and print a deprecation warning.

mintar:
Also note that `serial_name` was never used in the code, and the parameter was not changed. Only the wiki documentation was incorrect.

seebq:
Understood, thank you!

Feel free to delegate issues/fixes to me @seebq or @edavies64 on this library if needed, we'd like to help.

mintar:
> Feel free to delegate issues/fixes to me @seebq or @edavies64 on this library if needed, we'd like to help.

Thanks for your offer, I'll remember that!

PR #84, opened 2021-03-08
Log synchronization window details at DEBUG level

mintar:
The message is logged in quick succession at the callback rate until
synchronization succeeds.
This can lead to a lot of warning messages even though the
synchronization succeeds quickly.

The message is confusing for users that don't know the code context
(which is very well explained in the comments by the way).

Given that the message is logging an internal detail, I propose to
change the level to DEBUG.


mintar:
This is a forward port of #82 to foxy. Not tested, but change is minimal.

clalancette (approved):

PR #97, opened 2021-08-06
Make sure libphidget22 library can be found.

clalancette:
In Foxy and later, we need to provide the .dsv hook so that
the library can be found.

Still a draft since I want to make sure it actually fixes the reporters problem.  If this does fix the problem, then I'll forward port this to Galactic as well.

### DIFF
--- a/libphidget22/CMakeLists.txt
+++ b/libphidget22/CMakeLists.txt
@@ -39,5 +39,8 @@ install(
 )
 
 ament_environment_hooks(env_hook/libphidget22_library_path.sh)
+set(ENV_VAR_NAME "LD_LIBRARY_PATH")
+set(ENV_VAR_VALUE "opt/libphidget22/lib")
+ament_environment_hooks(env_hook/libphidget22_library_path.dsv.in)
 
 ament_package(CONFIG_EXTRAS "cmake/libphidget22-extras.cmake.in")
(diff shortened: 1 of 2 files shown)


libphidget22/CMakeLists.txt
```diff
@@ -39,5 +39,8 @@ install(
 )
 
 ament_environment_hooks(env_hook/libphidget22_library_path.sh)
+set(ENV_VAR_NAME "LD_LIBRARY_PATH")
+set(ENV_VAR_VALUE "opt/libphidget22/lib")
```
mintar, line 43: I haven't tested this, but isn't there a `"/"` missing before `opt`?

clalancette: > I haven't tested this, but isn't there a `"/"` missing before `opt`?

No, we don't actually install libphidget22 to `/opt`, we install it to a relative `opt` directory.  In the case of the debian packages, that ends up being `/opt/ros/<rosdistro>/opt/libphidget22/lib`.  In the case of a local build, that ends up being `ws/install/libphidget22/opt/libphidget22/lib`.  By making this a relative path, it should work in both circumstances.

mintar: Ok, thanks for the clarification!

mintar (approved):

PR #100, opened 2021-08-20
Make sure libphidget22 library can be found. (#97)

clalancette:
In Foxy and later, we need to provide the .dsv hook so that
the library can be found.

This is the same fix as #97, but for Galactic.


clalancette:
@mintar friendly ping for review here, should be a simple one.

mintar (approved):
LGTM!

clalancette:
Thanks for the review, merging.

PR #104, opened 2021-09-29
Fix typo in error message

mintar:


clalancette (approved):
We should also port this fix to the `foxy` branch

mintar:
Merged and ported to `foxy`.

PR #118, opened 2022-02-16
spatial: Fix behavior after USB reattachment

mintar:

### DIFF
--- a/phidgets_api/include/phidgets_api/spatial.h
+++ b/phidgets_api/include/phidgets_api/spatial.h
@@ -70,7 +70,7 @@ class Spatial final
     std::function<void(const double[3], const double[3], const double[3],
                        double)>
         data_handler_;
-    PhidgetSpatialHandle spatial_handle_;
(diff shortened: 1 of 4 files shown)


mintar:
@clalancette : If you approve this PR, I'll merge into noetic and rebase onto foxy + galactic. Do you prefer separate PRs for foxy + galactic, or should I just rebase + push directly?

clalancette (approved):
Looks good to me, thanks for looking into this.

> @clalancette : If you approve this PR, I'll merge into noetic and rebase onto foxy + galactic. Do you prefer separate PRs for foxy + galactic, or should I just rebase + push directly?

I prefer separate PRs.  It's slightly more annoying, but it gives an easier trail to follow if we need to do any archeology later.

mintar:
> I prefer separate PRs. It's slightly more annoying, but it gives an easier trail to follow if we need to do any archeology later.

Not annoying at all. From the command line, it's only a matter of seconds. Plus, all the CI tests are actually run. So here you go: #119, #120.

PR #119, opened 2022-02-16
[foxy] spatial: Fix behavior after USB reattachment

mintar:


mintar:
I'll look into the build failure tomorrow. Same for #120.

mintar:
Build failures have been fixed, #119 and #120 are ready for review now.

clalancette (approved):

PR #120, opened 2022-02-16
[galactic] spatial: Fix behavior after USB reattachment

mintar:

### DIFF
--- a/phidgets_api/include/phidgets_api/spatial.hpp
+++ b/phidgets_api/include/phidgets_api/spatial.hpp
@@ -47,7 +47,9 @@ class Spatial final
                      bool is_hub_port_device,
                      std::function<void(const double[3], const double[3],
                                         const double[3], double)>
-                         data_handler);
+                         data_handler,
+                     std::function<void()> attach_handler = nullptr,
+                     std::function<void()> detach_handler = nullptr);
 
     ~Spatial();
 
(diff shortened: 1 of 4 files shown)


clalancette (approved):

clalancette:
Approved both this one and the one for Foxy.

@mintar Once you've merged these in, I'll suggest doing a release, at least for Rolling.  That should fix the currently broken patch in the release repository.  Thanks!

mintar:
@clalancette wrote:

> @mintar Once you've merged these in, I'll suggest doing a release, at least for Rolling. That should fix the currently broken patch in the release repository. Thanks!

Yes, that was exactly my plan. I've been planning to do releases for foxy, galactic and rolling.

I've merged both PRs, did `catkin_generate_changelog` and `catkin_prepare_release`, but I'm stuck at the `bloom-release` step because I don't have write access to the https://github.com/ros2-gbp/phidgets_drivers-release repository (I've only done ROS1 releases so far). Could you give me access, please?

clalancette:
@mintar I've opened up https://github.com/ros2-gbp/ros2-gbp-github-org/pull/11 , which should restore your access once it is merged and deployed.  I'll let you know when that happens.

clalancette:
All right, you should have permissions now.  Let me know if you are still having trouble.

nuclearsandwich:
> All right, you should have permissions now. Let me know if you are still having trouble.

In point of fact, you have likely received an invitation to join the ros2-gbp organization which you'll need to accept in order to use the ros2-gbp release repository.

mintar:
Thanks! I've released phidgets_drivers into noetic, foxy, galactic and rolling now.

PR #126, opened 2022-03-22
BUGFIX: duplicated values for all encoders

mintar:


mintar:
The build failure on rolling is caused by warnings from the libphidget library and unrelated to this PR.

clalancette (approved):

PR #127, opened 2022-04-05
Noetic networkhub

J-C-Haley:
Added two parameters, server_name and server_ip, to allow connection to network & wireless VINT hubs to all libraries. Tested only on a temperature phidget, but should work on anything that connects to a vint hub.


phidgets_digital_outputs/launch/digital_outputs.launch
```diff
@@ -23,6 +23,12 @@
 
     # optional param is_hub_port_device, used if connected to a VINT hub
     <!-- <param name="is_hub_port_device" value="false"/> -->
+
+        # optional param server_name, required to locate remote network servers
+    <!-- <param name="server_name" value="myPhidgetServer" /> -->
```
mintar, line 28: There's some extra whitespace at the beginning of this line.

mintar (changes requested):
Apart from the tiny whitespace issue (see my comment above), this looks good to me.

mintar:
Thanks the PR, and for copy-pasting it everywhere. @clalancette : We should really think about whether we want to create a superclass where we can put all of this stuff so we don't duplicate code everywhere...

clalancette:
> Thanks the PR, and for copy-pasting it everywhere. @clalancette : We should really think about whether we want to create a superclass where we can put all of this stuff so we don't duplicate code everywhere...

That seems reasonable to me, though I likely won't have time to do it in the near future.  I'll suggest we go ahead with this PR as-is (with whatever changes it needs), and then we can do a separate PR just doing that cleanup.

mintar:
>  I'll suggest we go ahead with this PR as-is (with whatever changes it needs), and then we can do a separate PR just doing that cleanup.

Yes, I was going to suggest that as well.

mintar:
I've fixed the small whitespace error and merged the PR in 6734150. Thanks for your contribution!

J-C-Haley:
Thanks! Hurray for my first PR. My two cents is that the specifics for hardware diversification have to happen on some level, so it's kind of nice having these settings in all the repos individually? Then in a bigger launch file different phidgets can be all set with one argument

mintar:
It was your first PR? Nice! :)

J-C-Haley:
Thanks! And actually, I've just discovered setting the server parameters isn't necessary if one uses network discovery:

PhidgetNet_enableServerDiscovery(PHIDGETSERVER_DEVICEREMOTE);

would there be any reason to not have this enabled? I can add it as another launch parameter option, or we could just bake it in?

mintar:
I'd prefer it as a parameter.

PR #129, opened 2022-04-08
Add pre-commit, move from travis to GitHub actions, fix style

mintar:
See separate commits for detailed descriptions.


clalancette (approved):
Yes please!  This looks like a good thing to have; once we merge this, we should probably port over to the ROS 2 branches.

One thing is that I didn't see the workflow actually run in this PR.  I *thought* that was supposed to happen when you initially added a `.github/workflows` file, but I don't see it.  Any thoughts on what is going on there?  Regardless, I think this is OK to merge in, as we can always do a follow-up with fixes.

mintar:
I was wondering the same! Anyhow, now it's building for the merge commit (because it's a push): https://github.com/ros-drivers/phidgets_drivers/actions/runs/2116423918

Let's hope it will also build for future pull requests. Let's observe this.

PR #131, opened 2022-04-08
[foxy] Add pre-commit, move from travis to GitHub actions, fix style

mintar:


.github/workflows/github-actions.yml
```diff
@@ -0,0 +1,40 @@
+name: Build and run ROS tests
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        rosdistro: [foxy]
+      fail-fast: false
+    runs-on: ubuntu-latest
```
clalancette, line 9: I'm going to suggest that we explicitly call out ubuntu-20.04 here.  At some point in the future, it is likely that `ubuntu-latest` will map to ubuntu-22.04, and then everything will break.

clalancette (changes requested):
I left one thing to fix (thinking about it now, we may want to fix the same thing in Noetic).  Other than that, this looks good.

.github/workflows/github-actions.yml
mintar: I don't think it will break, because the `container:` tag selects a specific Docker image where everything runs. If you look at the tests (e.g., the apt-get part), you'll see that noetic, foxy and galactic actually run on Ubuntu 20.04 and rolling runs on 22.04. I don't really know what significance the`runs-on` has when you specify `container`.

mintar: I've re-read the documentation, and the `runs-on` only specifies which host system we want (in this case, Ubuntu and not Windows or Mac, but we don't care which Ubuntu version). So I think leaving it at `ubuntu-latest` is the right thing to do.

clalancette: Ah, makes sense!  Thanks for looking into it.  In that case, I'm happy with this and I'll approve.

clalancette (approved):

PR #134, opened 2022-04-08
sensor_msgs/Temperature and multi-channel readers

J-C-Haley:
Added a sensor_msgs/Temperature topic, giving published temperatures a timestamp. Prior float64 is preserved for now.

Added channel parameter to phidgets_temperature, allowing for multi-port readers.


phidgets_temperature/launch/temperature.launch
```diff
@@ -21,11 +21,15 @@
...
+    # optional param channel, defaults to 0, used if multiple reader ports are available
+    # Launch seperate nodelets for seperate channels 
+    <!-- <param name="channel" value="0"/> -->
+
     # optional param thermocouple_type, defaults is 0
     <!-- <param name="thermocouple_type" value="0" /> -->
 
     # supported data rates: 4 8 16 24 32 40 ... 1000 (in ms)
-    <param name="data_interval_ms" value="4"/>
+    <param name="data_interval_ms" value="250"/>
```
J-C-Haley, line 32: 4 ms doesn't work on a thermocouple reader, setting to default in readme

mintar:
Thanks! One of the checks is failing. This is because I added a check yesterday to ensure that the code formatting complies with clang-format. To fix this, follow the [pre-commit Formatting Checks](https://github.com/ros-drivers/phidgets_drivers#pre-commit-formatting-checks) section of the README. If you run `pre-commit run -a`, it should fix the formatting automatically for you. Once you commit that, the check should pass.

mintar (changes requested):
Thanks again for the PR. I've just reviewed it, and I think we should handle multiple channels the same way that they are handled in the other phidgets (for example, you can use the `digital_output` files as a template, reusing as much code as possible): One `temperature` class per channel, then one `temperatures` class to group all channels.

mintar:
Closing PR because author stopped responding.

J-C-Haley:
Apologies, this is fine, I imagine you've moved on.

PR #137, opened 2022-04-10
Support configuring encoder data interval and IO Mode

### LINKED ISSUE #122: Allow configuring encoder IOMode
The new encoder module ENC1000_0 has several [IOMode options](https://www.phidgets.com/docs/Encoder_Primer#Output_Circuit):

```cpp
typedef enum {
 ENCODER_IO_MODE_PUSH_PULL = 0x1,
 ENCODER_IO_MODE_LINE_DRIVER_2K2 = 0x2,
 ENCODER_IO_MODE_LINE_DRIVER_10K = 0x3,
 ENCODER_IO_MODE_OPEN_COLLECTOR_2K2 = 0x4,
 ENCODER_IO_MODE_OPEN_COLLECTOR_10K = 0x5,
} Phidget_EncoderIOMode;
```

This should be configurable from the node. I'd like to add it as a node parameter (and I'll send a PR), but I'd like to know if this should be configured as a string (and match it to the enum) or an integer and just pass it directly?

lights0123:
Closes #122

---

This was written as part of my employment at @RoadPrintz. I have received permission to license these changes under the relevant upstream licenses.

### DIFF
--- a/phidgets_api/include/phidgets_api/encoder.h
+++ b/phidgets_api/include/phidgets_api/encoder.h
@@ -80,7 +80,7 @@ class Encoder final
     void setIOMode(Phidget_EncoderIOMode io_mode) const;
 
     uint32_t getDataInterval() const;
-    void setDataInterval(uint32_t interval) const;
+    void setDataInterval(uint32_t data_interval_ms) const;
 
     void positionChangeHandler(int position_change, double time,
                                int index_triggered);
--- a/phidgets_api/include/phidgets_api/encoder.h
+++ b/phidgets_api/include/phidgets_api/encoder.h
@@ -76,6 +76,12 @@ class Encoder final
      * @param enabled The new powered state of the encoder*/
     void setEnabled(bool enabled) const;
 
+    Phidget_EncoderIOMode getIOMode() const;
+    void setIOMode(Phidget_EncoderIOMode io_mode) const;
+
+    uint32_t getDataInterval() const;
+    void setDataInterval(uint32_t interval) const;
+
     void positionChangeHandler(int position_change, double time,
                                int index_triggered);
 
--- a/phidgets_high_speed_encoder/include/phidgets_high_speed_encoder/high_speed_encoder_ros_i.h
+++ b/phidgets_high_speed_encoder/include/phidgets_high_speed_encoder/high_speed_encoder_ros_i.h
@@ -69,8 +69,6 @@ class HighSpeedEncoderRosI final
     // (Default=1) Number of "ITERATE" loops without any new encoder tick before
     // resetting the filtered average velocities.
     int speed_filter_idle_iter_loops_before_reset_;
-    Phidget_EncoderIOMode io_mode_;
-    int data_interval_;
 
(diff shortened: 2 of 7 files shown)


phidgets_api/include/phidgets_api/encoder.h
```diff
@@ -76,6 +76,12 @@ class Encoder final
      * @param enabled The new powered state of the encoder*/
     void setEnabled(bool enabled) const;
 
+    Phidget_EncoderIOMode getIOMode() const;
+    void setIOMode(Phidget_EncoderIOMode io_mode) const;
+
+    uint32_t getDataInterval() const;
+    void setDataInterval(uint32_t interval) const;
```
mintar: Could you rename `interval` to `data_interval_ms` or `interval_ms` for consistency with the rest of the code base? (Unfortunately the code base isn't quite consistent here either).

phidgets_high_speed_encoder/include/phidgets_high_speed_encoder/high_speed_encoder_ros_i.h
```diff
@@ -69,6 +69,8 @@ class HighSpeedEncoderRosI final
     // (Default=1) Number of "ITERATE" loops without any new encoder tick before
     // resetting the filtered average velocities.
     int speed_filter_idle_iter_loops_before_reset_;
+    Phidget_EncoderIOMode io_mode_;
+    int data_interval_;
```
mintar: These don't need to be member variables, right? (see phidgets_motors for example)

I don't have a strong opinion on this, we can leave it as member variables if you prefer. I just try to keep the diff between the different phidgets small because there's a lot of duplicated code, and at some point somebody should refactor it (introducing a superclass or something), which will be easier the fewer differences we have between phidgets.

phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
```diff
@@ -74,6 +74,45 @@ HighSpeedEncoderRosI::HighSpeedEncoderRosI(ros::NodeHandle nh,
     {
         publish_rate_ = 0;
     }
+    if (!nh_private.getParam("data_interval", data_interval_))
+    {
+        data_interval_ = 0;
```
mintar: So this is a deviation from the other phidgets, where we always explicitly set a default value. But actually I like your version better (leaving the device default). So no need to change anything here, just a remark.

mintar (changes requested):
Great, thanks! This looks very clean. All I have are a few minor suggestions (see comments).

phidgets_high_speed_encoder/src/high_speed_encoder_ros_i.cpp
mintar: Oh, one more thing: This parameter has to be renamed to `data_interval_ms`, because that's what all the other phidgets use.

mintar (approved):
LGTM!

mintar:
Thanks a lot!

PR #138, opened 2022-04-11
Update to libphidgets 20220203

### LINKED ISSUE #133: Fix gcc warnings on Ubuntu 22.04
The gcc version in Ubuntu 22.04 seems to have added a few more warnings compared to Ubuntu 20.04. That's why the builds on the buildfarm are now marked as "unstable", e.g. here:

https://build.ros2.org/job/Rpr__phidgets_drivers__ubuntu_jammy_amd64/2/gcc/

Somebody should fix those warnings in `libphidget22` and generate a new patch file (`libphidget/patch`). I don't have time for that though, and also perhaps it's better to leave the code as it is instead of "fixing" something to make the warnings go away and breaking something in the process. Or perhaps we should just silence a few more warnings in the CMakeLists.txt?

clalancette:

This should fix #133 , and also update us to the latest libphidget code from upstream.

I've targeted this at the `galactic` branch, but I actually think we should leave Galactic as it is and make a new, separate branch for Rolling.  Then I'll retarget this at that new branch.  @mintar what do you think?


mintar (approved):
LGTM! Also +1 on creating a new rolling branch. Or we could create a "humble" branch instead and release that to rolling and humble (same as we did with the galactic branch so far).

clalancette:
> Also +1 on creating a new rolling branch. Or we could create a "humble" branch instead and release that to rolling and humble (same as we did with the galactic branch so far).

What we've been finding is that with the advent of Rolling, it is better to have a separate branch for Rolling, and then a single branch for each released distribution.  It is a bit more work in terms of porting things back and forth, but it makes it much easier for everyone to understand what branch corresponds to what.

So my proposal is to create a `rolling` branch right now, and retarget this PR to it.  Once we officially branch off Humble (next week sometime), we can create a `humble` branch.  What do you think?

mintar:
Yes, sounds good! This is more or less how I'm doing it with the ROS1 branches in my repos as well. Let's do it!

mintar:
> So my proposal is to create a rolling branch right now, and retarget this PR to it. Once we officially branch off Humble (next week sometime), we can create a humble branch.

Now that humble is released and there are Docker images for it, we should do this.

* Create a `humble` branch that we keep synced to the `rolling` branch for now.
* Update the `github-actions.yml` build matrix to `rosdistro: [humble, rolling]`.
* Update the rosdep `humble/distribution.yaml` to point to the `humble` branch.

Agreed?

@clalancette : Since you're the ROS2 maintainer of this repo, I'll leave this to you. Unless you want me to do it.

clalancette:
> @clalancette : Since you're the ROS2 maintainer of this repo, I'll leave this to you. Unless you want me to do it.

I'll go ahead and make the branch, though I'm not entirely sure we should keep them synced and using the same matrix.  That is, it would be simpler to just have the humble branch test against humble, and the rolling branch against rolling.  It requires a bit more work to backport everything, but I think it is much simpler for users to understand.

Regardless, I'll create a humble branch (based off of rolling), then open a PR where we can discuss further.

PR #140, opened 2022-04-13
Switch to only building for Rolling.

clalancette:


mintar (approved):
LGTM! Once humble is branched off we have to remember to add it too.

PR #149, opened 2022-06-20
Add in humble to the test matrix.

clalancette:
That way we will test incoming PRs on both Rolling and Humble.
We'll maintain this situation until we run into a situation
where Rolling and Humble need to diverge.  This commit
should be fast-forwarded onto the humble branch.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


mintar (approved):

mintar:
LGTM!

clalancette:
Great, thanks.  I'm going to go ahead and merge this and the fast-forward it to the Humble branch.

PR #155, opened 2022-12-21
BUGFIX: Wrong speed conversion factor [phidgets_high_speed_encoder]

jlblancoc:
The current code assumed time intervals from the Phisgets API was microseconds, but it's actually milliseconds. Reported speeds are all wrong by a factor of 1e3.


mintar:
Thank you!

jlblancoc:
You're welcome! There's one more bug whose fix will come shortly....

PR #156, opened 2022-12-22
BUGFIX: Wrong speed conversion factor

mintar:
The current code assumed time intervals from the Phisgets API was microseconds, but it's actually milliseconds. Reported speeds are all wrong by a factor of 1e3.

This is a forward-port of #155 to ROS2 galactic.


mintar:
Closing because galactic is EOL.

PR #158, opened 2022-12-23
BUGFIX: Z-channel index was not observed in reported positions

jlblancoc:
For encoders with the Z-channel correctly wired to the Phidgets board, the incremental encoder positions were reported without taking into account the passes through zero, effectively ignoring the Z-channel effect.

This only affected positions, not velocities.

PS: As usual, the same patch should be applied to all branches...


mintar:
Thanks!

mintar:
Cherry-picked onto noetic, merged into humble and rolling.

PR #160, opened 2023-02-28
Update to libphidget22 1.13.

clalancette:
Here's my once-a-year update to the latest libphidgets22.  This time, we update to 1.13.20230224, which just came out a few days ago.

Note that I no longer have access to phidgets, so I can't test this anymore.  @mintar I'll leave it to you to decide whether to take this update or not.


mintar:
Just tested it with my Phidgets Spatial, works great. Thanks!

PR #168, opened 2023-11-10
Port PR #153 (add support for onboard orientation estimation) to ROS2

mintar:

PR #169, opened 2023-11-24
Only set magnetometer gain if param is set

mintar:
This is required because the function `PhidgetSpatial_setAlgorithmMagnetometerGain` is present in the Phidgets Spatial MOT0109 API (https://www.phidgets.com/?prodid=1204#Tab_API), but was removed in the MOT0110 API (https://www.phidgets.com/?prodid=1205#Tab_API). This caused an exception to be thrown on the MOT0110 (https://github.com/ros-drivers/phidgets_drivers/pull/168#issuecomment-1823622125).

This commit allows not setting the parameter on the MOT0110 and onwards, thereby avoiding the function call and the exception.

### DIFF
--- a/phidgets_spatial/launch/spatial-launch.py
+++ b/phidgets_spatial/launch/spatial-launch.py
@@ -38,7 +38,8 @@ def generate_launch_description():
         'ahrs_bias_time': 1.25,
 
         # optional param algorithm_magnetometer_gain, default is 0.005
-        'algorithm_magnetometer_gain': 0.005,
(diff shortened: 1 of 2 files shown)


samerhanoun:
Works now. Orientation data is in agreement with imu_filter_madgwick.
launch
topic list
topic echo
compare phidgets_drivers with madgwick

Happy to test any further fixes if needed, for now we have the orientation data which we needed. Thanks for your help.

clalancette (approved):

PR #172, opened 2024-02-16
Add support for VINT networkhub to ROS2

### LINKED ISSUE #135: Port #127 to ROS2 (Add support for VINT networkhub)
See #127

mintar:
This is a port of #127 to ROS2.

Closes #135.

### DIFF
--- a/phidgets_temperature/src/temperature_ros_i.cpp
+++ b/phidgets_temperature/src/temperature_ros_i.cpp
@@ -63,6 +63,20 @@ TemperatureRosI::TemperatureRosI(const rclcpp::NodeOptions& options)
         throw std::runtime_error("Publish rate must be <= 1000");
     }
 
+    this->declare_parameter("server_name",
+                            rclcpp::ParameterType::PARAMETER_STRING);
+    this->declare_parameter("server_ip",
+                            rclcpp::ParameterType::PARAMETER_STRING);
+    if (this->get_parameter("server_name", server_name_) &&
+        this->get_parameter("server_ip", server_ip_))
+    {
+        PhidgetNet_addServer(server_name_.c_str(), server_ip_.c_str(), 5661, "",
+                             0);
+
+        RCLCPP_INFO(get_logger(), "Using phidget server %s at IP %s",
+                    server_name_.c_str(), server_ip_.c_str());
+    }
+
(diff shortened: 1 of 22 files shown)


mintar:
I could not test this PR because I don't have access to a VINT network hub. @patilunmesh, could you test this please?

phidgets_temperature/src/temperature_ros_i.cpp
```diff
@@ -63,6 +63,20 @@ TemperatureRosI::TemperatureRosI(const rclcpp::NodeOptions& options)
         throw std::runtime_error("Publish rate must be <= 1000");
     }
 
+    this->declare_parameter("server_name",
+                            rclcpp::ParameterType::PARAMETER_STRING);
+    this->declare_parameter("server_ip",
+                            rclcpp::ParameterType::PARAMETER_STRING);
+    if (this->get_parameter("server_name", server_name_) &&
+        this->get_parameter("server_ip", server_ip_))
```
clalancette, line 71: I don't know much about the `PhidgetNet_addServer` API.  If you call it to set the server name, does it always take precedence over the `serial_num`/`hub_port` that is otherwise specified?

If that is the case, then I think we should print a warning or error if the serial_num/hub_port is specified to anything but the default.  Also, I think that the `RCLCPP_INFO` call below should probably conditionally print whether it is attaching to the server, or to the hub_port, depending on which one is configured.  @mintar what do you think?

mintar: I must admit I don't know the first thing about either the VINT serial hub nor the VINT network hub. I don't have either device for testing; that's what's making this so challenging even though it's really simple code.

In this PR, I've just ported the working code from #127 to ROS2. I haven't changed anything except replacing the ROS1 API calls with their ROS2 equivalents. Since we have this code in Noetic for a while now, and nobody has complained, I expect it's working.

You're definitely right about the RCLCPP_INFO call, but I believe somebody with access to a device for testing should make those changes.

patilunmesh:
@mintar I tested the code with a VINT hub using a temperature, humidity phidget and it is working!!

mintar:
@clalancette: Seems to be working as in Noetic. Do you think this is ready for merging?

mintar:
@clalancette: Can I merge this?

PR #175, opened 2024-03-11
Update to libphidget22 1.19

mintar:
This version of libphidget22 compiles without warnings, so we don't need the patch file any more.


mintar:
@clalancette : What do you think about this PR? Upgrading the libphidget22 from time to time is necessary to support new devices.

If you agree to this PR, I'll open another one for the ROS2 branches (humble, iron + rolling).

clalancette (approved):
Yep, totally on-board with updating this.

PR #176, opened 2024-03-13
Update to libphidget22 1.19

mintar:
This version of libphidget22 compiles without warnings, so we don't need the patch file any more.


mintar:
This is a port of #175 to ROS2.

mintar:
There seem to be two different build errors:

1. [ROS build farm](https://build.ros2.org/job/Rpr__phidgets_drivers__ubuntu_noble_amd64/2/console):

    ```
    13:18:23 Step 26/35 : RUN pip3 install -U setuptools==59.6.0
    13:18:23  ---> Running in 400d80ac7b16
    13:18:23 [91merror: externally-managed-environment
    ```

2. [GitHub actions for Rolling](https://github.com/ros-drivers/phidgets_drivers/actions/runs/8264474799/job/22608150710?pr=176):

```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
phidgets_gyroscope: No definition of [ament_cmake_ros] for OS version [jammy]
```

AFAICT:

re. 1.: The build farm is just acting up again, the error is unrelated and should be fixed when the build farm is fixed.

re. 2.: Seems like the rolling rosdep support for Ubuntu 22.04 has been removed; rolling has probably already rolled over to Ubuntu Noble 24.04? That's a bit of a shame because 24.04 is not released yet, and consequently GitHub actions doesn't provide it yet.

Since this PR builds successfully on humble and iron, I suggest we merge this into humble, iron and rolling, and after Ubuntu 24.04 becomes available for GitHub we fix that build error in a different PR.

@clalancette: Any thoughts?

clalancette:
> re. 1.: The build farm is just acting up again, the error is unrelated and should be fixed when the build farm is fixed.
> 
> re. 2.: Seems like the rolling rosdep support for Ubuntu 22.04 has been removed; rolling has probably already rolled over to Ubuntu Noble 24.04? That's a bit of a shame because 24.04 is not released yet, and consequently GitHub actions doesn't provide it yet.

Yes, this is exactly right; all of this has to do with the transition to Ubuntu 24.04, which we are still in the midst of.  Given that Humble and Iron are passing, I'm fine with merging this as-is.

clalancette (approved):

PR #182, opened 2024-04-26
Set CMake policy CMP0135

mintar:
This hopefully fixes the following error on ROS2 Rolling (Ubuntu 24.04):

    The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is...

See https://build.ros2.org/job/Rdev__phidgets_drivers__ubuntu_noble_amd64/4/cmake/


mintar:
Never mind, I'm going to close this and leave the warning unfixed, because Ubuntu 22.04 (ROS2 Iron) has cmake 3.22, but CMP0135 was introduced in cmake 3.24. Therefore, we cannot merge this PR into the `iron`/`humble` branches, and splitting off `rolling` just for this warning isn't worth it.

PR #187, opened 2025-05-07
CI: Update test matrix

mintar:
Dropping `iron` because it's end-of-life (so the rosdep keys don't resolve any more), and adding `jazzy`.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


mintar:
@ros-pull-request-builder retest this please

mintar:
The GitHub action checks for humble, jazzy, and rolling all pass.

The Jenkins build farm checks for kilted and rolling fail (for unrelated reasons: the buildfarm fails to clone the Git repo), jazzy passes, humble is not tested on "rolling" branch.

Merging.

PR #188, opened 2025-05-08
Add zeroing service to high_speed_encoders

cedricpradalier:
This PR adds a zeroing service to high_speed_encoders. 
The service uses a trigger message, adapted from std_srvs, and potentially reusable in other nodes.

The block of INFO logs is moved where the str variable makes sense (it was redefined on line 120). 

Note: I tried implementing the zero with setPosition but, for some reason, even if the change was successful within the service call (validated with RCLCPP_INFO), the value was lost in the main thread. As a result, I had to add the absolute_zero_ variable to store the zero.


mintar:
Very nice, thank you! Next time, starting with a fresh branch for each pull request is best. That way, only the commits that are actually part of the pull request are shown here, not the ones that are already merged into humble. It doesn't really matter in this case because I'm going to squash the commits anyway, and because I happen to know which commits are actually having an effect (51f87eb, 7827542, 754d65e, and 61aae29).

PR #191, opened 2025-05-25
[kilted] Update deprecated call to ament_target_dependencies

mintar:


mintar:
@clalancette : Do you have time to review this PR? It fixes the builds for rolling and removes the deprecation warning on kilted. It's best to review each commit separately.

I'm pretty sure this PR is okay. Maybe have a look at 6b0a6c74 to see if this is how a standard ROS2 CMakeLists.txt should look. If you don't have time that's okay, I'll just merge it in a few days then (unless you object).

I plan to merge this into `rolling` and then manually fast-forward `jazzy` and `kilted` to the same commit. The checks show that it builds on all 3 distros.

It doesn't build on `humble` however, so I suggest we just leave `humble` where it is and no longer make new releases for humble.

Since this PR fixes the failing builds on rolling, I plan on merging this soon.

mintar:
@clalancette: I'm going to merge this on Friday unless you have any objections.

PR #194, opened 2025-07-10
Remove unnecessary linking to component_manager.

clalancette:
These libraries are components themselves, and never need any parts of component_manager.  So just remove this dependency.

@mintar This is a follow-up to https://github.com/ros-drivers/phidgets_drivers/pull/191 (apologies that I did not respond at the time you opened it).

On a separate but related note, given that Noetic is now End-of-Life, would you mind if we made the default branch for this repository `rolling`?


mintar:
Thanks!

> On a separate but related note, given that Noetic is now End-of-Life, would you mind if we made the default branch for this repository rolling?

Done!

clalancette:
Thanks, much appreciated.

